PR #20, opened 2019-11-15
Massively Parallelize MediaScanner and thread safety

randomnicode:
This PR is built on top of JSR310.

MediaScanner is currently a single-threaded low-priority operation full of object level pessimistic synchronization locking. This PR attempts to massively parallelize it and additionally alleviate some of the object locks to significantly boost throughput.

- Add thread safety for touching objects because now they are subject to massive number of concurrent threads
- Uses a custom forkjoin pool to fork tasks. This keeps the commonPool free and allows work-stealing to distribute load across all cores
  - Parallelism can be specified by environment variable `MEDIASCANNER_PARALLELISM`
  - Default parallelism is `number of core + 1`
- All FJP threads retain their low priority status, so it should have minimal impact on the system

I'll annotate the PR to help follow through. Feel free to ask questions or point out if you see an issue.

This PR should make a considerable bit of difference for slow media scans. My systems see a fairly large boost. Most systems are nowadays multi-core. We can likely optimize even more once Spring 2 and Hikari are merged because I suspect now the bottleneck will be the DB threadpool and we can tweak to increase that threadpool size.

The only caveat here is the DB connection. If it's single-threaded, then you're just effectively making it a single-threaded operation by bottlenecking it at the db connection level. You need a connection pool to fully see the benefits.

**Benchmarks:**
For a miniscule amount of data: (about 3295 entries that I could conjure up for a test bench) on a new internal db for both test running in serial on the same 6 core, 12 SMT (which would make parallelism 13 for the latter test by default), I see about an 800% performance boost, without any optimizations.
- Old single-threaded scan: ~50s
```
2019-11-06 17:39:35.440  INFO --- o.a.p.s.MediaScannerService              : Starting to scan media library.
...
2019-11-06 17:40:24.378  INFO --- o.a.p.s.MediaScannerService              : Completed media library scan.
2019-11-06 17:40:24.909  INFO --- o.a.p.service.PlaylistService            : Starting playlist import.
2019-11-06 17:40:24.909  INFO --- o.a.p.service.PlaylistService            : Completed playlist import.
```

- Newer multithreaded parallel scan: ~6s
```
2019-11-06 17:43:13.753  INFO --- o.a.p.s.MediaScannerService              : Starting to scan media library.
...
2019-11-06 17:43:18.065  INFO --- o.a.p.s.MediaScannerService              : Completed media library scan.
2019-11-06 17:43:20.393  INFO --- o.a.p.s.MediaScannerService              : Media library scan took 6s
2019-11-06 17:43:20.399  INFO --- o.a.p.service.PlaylistService            : Starting playlist import.
2019-11-06 17:43:20.399  INFO --- o.a.p.service.PlaylistService            : Completed playlist import.
```

Mostly same PR as https://github.com/airsonic/airsonic/pull/1385


airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -191,7 +188,8 @@ private void doScanLibrary() {
             // Maps from artist name to album count.
             Map<String, AtomicInteger> albumCount = new ConcurrentHashMap<>();
             Map<String, Artist> artists = new ConcurrentHashMap<>();
-            Map<MediaFile, Album> albums = new ConcurrentHashMap<>();
+            Map<String, Album> albums = new ConcurrentHashMap<>();
```
eharris: What was the impetus for changing this from using `MediaFile`?

randomnicode: That's a good question. I'll try explaining from the beginning.

- The reason these maps were put in here in the first place is to accumulate data "in memory" so we can put/commit everything to the db together at the END of the media scanning process. Going back and forth to the DB DURING the media scanning process seemed like an expensive proposition to me, especially sometimes for the same data, given the latencies involved with a roundtrip to the DB.
- My reasoning formed when I saw this pattern happening:
  - Encounter a file
  - Fetch or create a row for it from the db
  - change some attribute for it in the db
  - a different method: fetch or create a row for it from the db (which has no idea about a different method doing the same thing, possibly in parallel)
  - change some attribute for it in the db
  - and so on
- I thought going to the db back and forth for every operation (sometimes changing the same thing at times) DURING scanning was an expensive proposition, and it might be helpful if we had an "in-memory db" that we just add to or update or take from while scanning and then commit the final product to the db in whole at the end.
- So, imagine we're scanning Artist1's Album1, and it has Song1, Song2, and Song3. And let's say this artist/album is the first time ever encoutered by the media scanner
  - Previously:
    - Encounter Song1: Create Artist1 in the DB, fetch that row, update its timestamp and recommit to DB. Create Album1 in the DB, update its timestamp, recommit. Create Song1.
    - Encounter Song2: Create Artist1 in the DB (already there), update its timestamp. Create Album1 (already there), update its timestamp. Create Song2.
    - Encounter Song3: Create Artist1 in the DB (already there), update its timestamp. Create Album1 (already there), update its timestamp. Create Song3.
  - Now:
    - Encounter Song1: Create Artist1 in the Map if not exists (doesn't), fetch that entry, update its timestamp. Create Album1 in the Map, update its timestamp. Create Song1.
    - Encounter Song2: Create Artist1 in the Map if not exists (it does, so its a no-op), update its timestamp. Create Album1 in the Map (noop), update its timestamp. Create Song2.
    - Encounter Song3: Create Artist1 in the Map if not exists (it does, so its a no-op), update its timestamp. Create Album1 in the Map (noop), update its timestamp. Create Song3.
    - At the end, commit the final Artist1 result, the final Album1 result to the DB. We've successfully minimized the churn DURING processing to in-memory structures instead of going back and forth with the DB, and the DB only sees the final product. We pay the penalty for going to the DB only once instead of of like 9 times every time we need to update (or create) something.
- So now we need to hold the data in memory, we need a key to address it with, during the in-processing and at the end during committing. This should be whatever the primary key in the db was (artist's name etc), or some other unique key they were using to retrieve and update the data in memory. For albums I made a mistake using a MediaFile as the key. My thought process was that it would create a MediaFile for an album, and then retrieve and update it on subsequent accesses, but what I forgot was that each song has a different MediaFile. So here's what ends up happening:
  - Album1 folder encountered. MediaFile1 created for Album1. Check map for album under MediaFile1 key, none found. Create album for Album1 and store in the Map with MediaFile1 as key.
  - Album1/Song1 encountered. MediaFile2 for Song1 created. Check map for album under MediaFile2 key, none found. Create album for Album1 and store in the Map with MediaFile2 as key.
  - Continue for all other songs. At the end, our map contains the same album split across (lets say 10 songs + 1 album), 11 keys, each with partial info. Instead of aggregating into one album, each song doesn't find the album and so creates one. And when we finally go to commit to the db, the DB starts throwing unique key violations. It'll create the first key's album, then will encounter the same album under a different key, and will say, I already have this album, cannot commit, and so forth.
- Thus this is the solution, the String key of the album map reflects the primary key of the album in the db (artist name/album name composite), and is generated here: https://github.com/airsonic-advanced/airsonic-advanced/pull/20/files#diff-44b0964f8fb27f359ad54d7377ef98b7R333-R334

eharris: Using a composite key of artist,album for albums doesn't seem like a good thing to me, as it will combine anything with matching artist and track names into the same "album".  This could be problematic in cases like someone has multiple rips of the same album in different formats and these would all be crammed together and make them painful to use.  Or if a bunch of different untagged albums are are labeled with generic "Unknown" values, these would also all end up mixed together in one hard to use "album".  Also, it seems like that might require additional protection code if the media files in the same "album" are spread across different media folders.

It would seem safer and better (to me) to use the deepest directory path (relative to the current top-level media folder) that contains the media file as the album primary key.  I specify that it's a relative path so that if media folders are moved but the structure below them is maintained, the database will find the moved files in the same "place", so the files won't need anything special to preserve ratings, playcounts, or other useful data.

randomnicode: Hmm, I think you're conflating MediaFiles (a separate table) with Albums. Whether I move the media files or not should have no bearing on the albums. And perhaps points to the existing (I didn't implement it) underlying design decision here (paths don't matter to albums).

Maybe it wasn't clear what I wrote. Whether it's a good idea or not doesn't matter, we're mandated by the current set up of the DB. You _have_ to use the composite key of (artist,album) because that's what the _unique key constraint_ of the album table in the database is (_NOT_ the path). Unless I change the table structure itself, whatever I use as a key for the in-memory map doesn't matter because once I try persisting in the db, that's the only determining factor.

Let's consider your scenario of album rips in different formats. Let's adopt your solution and say I use different keys for each in the map. At the end of the scan, this is how it's aggregated in the map (which is how you'd like it to be):
key1 -> Album1 (formatA) (Artist1), in some path
key2 -> Album1 (formatB) (Artist1), in a different path

Now lets try committing it to db:
- key1 -> Persists a row in the table with Artist1, Album1 (doesn't matter what the path is, though that's persisted too)
- key2 -> TRIES to persist a row in the table with the same Artist1, Album1, but immediately throws a unique key violation because a row with that composite key already exists. It doesn't matter that another column (the path) is different in this key's row

See the issue? For reference here's the key statement from the DDL of the albums table:
```
CREATE UNIQUE INDEX IDX_ALBUM_ARTIST_NAME ON ALBUM (ARTIST,NAME);
```
There is nothing, not even an index on the `PATH` column. So creating keys based on the path will undoubtedly lead to unique key violations when you try and persist in the database, because two paths with the same album name and artist will undoubtedly conflict in the DB.

Now of course we can redesign the table itself, but that'll be a different undertaking. I'm trying to parallelize the intake, not change the table structure itself. Currently the db as it's set up aggregates into the same albums no matter the path (and maybe that's by design). That's exactly also what the current airsonic installations do. Have you tried seeing how multiple rips of the same album look in your db table? Or how the unknowns are accumulated? They're all aggregated together.

Additionally, there is a flaw with using paths as a unique key: it implies a certain folder structure, which not everyone has. Songs are not always under an albums folder (some of my installations actually have all music under a "Music" folder barebones), I've seen others that rely on a genre-based segregation, and so on. I would not want airsonic to aggregate it all under the Music album. I would want it to read the metadata and aggregate it based on the metadata. But of course that has a different problem as you point out. Maybe adding a format key as part of the uniqueness constraint could be a solution, but then there is a further matter of a rip with half the songs in one format and the other in another...

So in short, redesigning albums needs to be handled with care, but isn't the concern of this PR. It just retains the existing infrastructure.

randomnicode:
If there are no more issues here, can I try and merge this?

eharris:
I guess I misunderstood then, I though you were also making a change to the form of the primary key in the database.

I'm not sure I have a complete handle on the entire change set, but I haven't noticed anything obviously wrong, so no objection to it being merged.

randomnicode:
No, I didn't touch any of the table DDLs (it would've been a file in liquibase/**)

eharris:
I do have one additional concern though, which is that won't storing all the scanning results in memory dramatically increase the memory requirements?  Maybe it's a worthy tradeoff, but I do know that people are running it in fairly memory constrained devices, like RaspberryPi's, which often also have very limited virtual memory capabilities (slow SD cards).

randomnicode:
I actually thought about that, which is why my first commit didn't have it. But then, I just don't think the amount of data is large enough.

It should have a minor effect, but it'll page memory if need be. Given the DB is residual on the filesystem anyway, retrieving from the slow SD cards again and again vs swapping memory on the card should make at worst no difference. On the positive side, if no paging is required, it'll boost performance.

eharris:
Except on smaller systems (again, I'll use a RPi as an example) the system disk (where swap exists) is often on slow SD flash (throughput of around 2 to 4MB/sec), while the data directories for airsonic are likely on a USB hard drive with considerably better throughput.

Isn't liquibase essentially a run-from-memory-but-persist-to-disk type of implementation anyway?  In that case, you're doubling the amount of memory needed since Airsonic now has its own copy of the entire db.  Add the Java VM and it's memory use, and everything is already quite memory heavy.  And paging to disk is all fine if you have enough swap, but with a large db plus java plus other things, adding significantly to the memory footprint of Airsonic seems a bit dangerous.

From my personal experience, I have 8GB ram in the machine I run airsonic on, and it still sometimes goes into swap hell and things go unresponsive for a while, and I know this is caused specifically by Airsonic/Tomcat/Java VM.  When I think of machines with only 256MB or 512MB of RAM, doubling memory use just isn't reasonable.

randomnicode:
Hmmm, not sure what liquibase has to do with this. Liquibase is merely a migration tool, it's only run at the very beginning at application startup, and isn't used at all thereafter.

I think you meant HSQLDB? In that case, not quite. While HSQLDB does have an in memory db, we use the file based one. While DB like HSQLDB does use memory (of course), it flushes to the disk asap, otherwise it'll lose whatever is in volatile memory even after we've committed. Imagine committing a transaction, losing power and then coming back up not to find any record of it because the DB forgot to flush its memory buffer to persistent storage.

Second, I think you're forgetting that a "higher" memory usage will only happen temporarily, and only at scan time which is, at most, once a day. You shouldn't really conflate it with the hiccups your machine is having randomly without looking at the timings of your swap hell and try and trace the hotspots. It happens to me too occasionally, but not during scan time. I suspect the root cause for that is some buffered stream that's not closing properly leading to a resource leak (completely speculating here).

Third, what do you think happens when we are retrieving these objects multiple times from the database? We have multiple copies of the same row in memory. Some of it will get garbage collected, some of it will get swapped. Either way, our memory footprint is no worse. We retrieve the DB into our memory anyway.

Last, I just don't think the memory footprint argument is strong enough. A few thousand objects on the heap isn't a big deal. I'll do some baseline testing to see the memory footprint increase, and whether my suspicion holds up.

If we're saying that the memory swap file and the db reside on two separate disks, I don't know what to say. We could likely ask folks to switch the swap to the same disk as the db?

eharris:
Yes, sorry, I meant HSQLDB.
No, I realized the higher memory use would only temporary, during a scan.  Doesn't matter how temporary it is if the host machine doesn't have enough memory to satisfy the needs.  Do we just say sorry, you are SOL and can't use Airsonic unless you get a bigger machine to run it?
So we have a few copies of some 1000 rows in memory at a time.  That's vastly different than hundreds of thousands of rows or multiple millions and multiple tables all the way up to (unlimited) for a very large library.
USB connected disks often take a while to come online and be detected, or are spun down when unused for a while, so swap is normally put on the primary boot volume rather than on application/intermittent storage devices.

randomnicode:
So my test benchmarks just confirm my hypothesis. The difference in memory is barely registrable.

For about a quarter of a million files, both applications (the current master vs master+PR) running on separate VMs, take up about the same amount of memory (about a 70 MB mean difference, which is well within the variance of multiple startups), with multiple test runs (~100) averaged. The variability across tests for each jar is actually greater than compared to each other.

I also did some common sense testing for lower number of files (like 10000), and the results are the same, with the nonparallelized version actually _losing_, though well within the variance of error again (171MB+/-69MB vs 164MB+/-31MB)

I don't think it's as straight-forward as "so you have a few copies of some 1000 rows in memory". Yes you do, but then you also need to garbage collect them. For massive libraries running on tiny amounts of memory, you can probably take your pick between GC pauses and massive memory churn vs utilizing (sometimes slower) swap space.

So, in short, based on the test benchmarks, I think this will be okay for most people. If there are reservations and folks who complain, we can address that in a follow up PR where we can replace the Maps with a LoadingCache which holds as much as it can in memory and writes to the DB upon eviction. Does that sound like a reasonable compromise?

eharris:
It's interesting (and a little funny) that you mention 70MB difference, as I looked and that is roughly the size of my on-disk database as well, so it seems reasonable that that is a good ballpark figure.

Your point about GC vs memory/swap churn is well taken.  I just wanted to note that increased memory might be a concern, as I hadn't done testing to verify it wasn't.  I do appreciate that you have.  And as far as your proposed compromise if we later find cases where increased memory use is a problem, I think that is a reasonably good path forward.  So no further objections to this being merged.

PR #28, opened 2019-11-18
Don't use batching for certain queries and remove extraneous queries

randomnicode:


randomnicode:
Just saw this comment. Sorry I merged it before.

Hmm, given that it's happening in a batch anyway, why is there a concern that the db will be blocked? Modern DBs are multithreaded, and can service multiple requests simultaneously, so one thread being busy shouldn't be a concern. Moreover, DBs are usually more efficient at optimizing a query than having an outsider try and optimize a query for them (such as artificially split things up in batches). Deleting stuff in batches is likely to keep the db thread hosed for longer (each time the db likely needs to seek the location of the data it needs to delete and thus pay a penalty, instead of just seeking once).

Is the question that another request might at least get to and be serviced by the DB between batches? Whereas with one request, the db connection is hosed until it returns? I don't think that's true with connection pooling...

Of course the best proof would be actually someone who has a large library. Do you have one and want to try it out and see if that's actually what's happening? If not, that's fine too. My theory is that it won't. If I'm wrong and someone complains, we can revert this commit. Until then I'd rather do things the proper DB query way and have the DB optimize the query.

eharris:
The concern is that any database that uses table-level locking for updates rather than row-level locking will block any other requests (even read requests) until the query (or transaction) is completed.  That is normally why inserts/updates/deletes are segmented into batches, to avoid locking out other users for long periods.

Even in cases where the db engine does support row-level locking, in some cases doing it as one big update can often cause unacceptable performance delays for other users, even if it doesn't completely lock them out.

randomnicode:
Hmm, these are good points, though I do think we need some field testing.

Table locking is indeed a concern and batches will indeed aid with table availability. However, and this is the key point, there is an index on the `present` column that is being used in the where clauses. As such, my impression is that we're worrying for not too much. I really don't think the amount of time it'll take to delete or update is a concern given the index present. Unless someone has 10s of millions of rows (doubtful).

The churn on the disk should be minimal given the index. Most times, a delete operation will be put in the work log and the previous row record will be marked deprecated. At vacuum (or checkpoint) time it should all be consolidated. So at worst we're looking at a bigger log file.

Finally, there is a bug in the original query:
```
        for (int id = minId; id <= maxId; id += batchSize) {	
            update(id to id+batchSize);
        }
```
Let's say minId is 500, maxId is 2000, and batch size is 1000
batch1: 500-1500
batch2: 1500-2500

see the issue? we're updating/deleting ids from 2000 to 2500 even though that's beyond maxId

eharris:
I would just be very uncomfortable removing the batching without having a LOT of widespread testing on it first.  Now I would be just fine with making the batch size a configuration option that even allows setting to "0" to disable batching.  Ship it with the prior value of 1000 being default, and allow people to play with different settings and report what works best.

eharris:
Also, indexes do speed up small updates as long as they are used in the query plan.  But an index will also often slow down overall throughput for mass updates because the index also has to be updated (more disk access, sorting, etc).  And you also are talking about optimizations that only make sense for databases that use a transaction log prior to updating the actual rows, and there are many that don't.

randomnicode:
I don't think anyone will play around with these values, so making them configurable wouldn't really have much point. We can revert this, but then we do need to fix the bug.

I pointed out that the operative clause in the query is indexed, so unless the table is tiny or there is an easier alternative, the query planner will undoubtedly use the index if its needed. Indices slow down writes in general, _but_ I've only seen those kinds of factors considered in massive DB updates (like above few million rows accounting but still sparse for the table). In those cases, doing things in batches is actually not done and the more ideal way is to remove indices, create new staging tables, truncate etc, none of which is done here because I just don't think that is the scope of the data involved.

Regarding tx logs, it was a minor point about side effects (in the sense that the original record is actually not usually touched, so disk churns etc are minimized, at the expense of a log file growth). Most DBs we support do use them. HSQLDB primarily (Postgres, SQL Server, Oracle etc.)

eharris:
I would certainly like to play with the batch size and do some testing.  And I think we need more than just a few developers to test this out, because there's lots of different hardware in use out there, so having a setting that non-developers can easily use to test and give feedback would seem to be a valuable thing to me.  And I think the ability to tune it based on the hardware in use could also be useful to users.  We already have a settings page for the database, why not just add the batch size as an option there?

And yes, we should fix the bug if it is a problem (I haven't had a chance to look at the code in enough depth to see yet, maybe there's a reason it hasn't mattered).

In any case, I am against removing the capability to limit the size of database updates.  I've personally had too many experiences in the past where not doing something like that to limit database impact caused serious performance problems for other database users.

eharris:
At a minimum, having a setting for batch size would allow us to change the default to no batching to see if any issues are encountered, but have it very easy for users to be able to change back to batching if they do see problems, no code changes or different version re-deployment necessary.  If nobody encounters problems after many people have lots of time on it, then sure, remove the capability altogether.

randomnicode:
Okay, you've convinced me, I think that's a pretty good idea. I'll push up a PR if you don't beat me to it.

Any preference for the name of the bulk update batch number variable? I'll just make it an env var for now and we can add jsp support for it whenever we want (unless I have time to do it all in one PR).

eharris:
I always hate having to name things, but something like DatabaseConfigUpdateRowLimit seems reasonable to me.

eharris:
For a cleaner history, it might be worthwhile to revert this commit and then apply a new PR to the original, just a suggestion.

PR #31, opened 2019-11-29
Batch updates with a specified row liimit

randomnicode:
Successor to https://github.com/airsonic-advanced/airsonic-advanced/pull/28 per discussion in that PR thread.

Exposes a custom specifiable db update row limit that may be used to control how the db is updated.

A value of 0 or below turns off the batching
A non negative, non zero value updates the db in batches of that size.

Higher value will attempt to update more rows in one roundtrip (thus requiring less roundtrips overall), but might lock the table for longer for each trip.
Lower value will update less rows in one roundtrip, thus locking the table continuously for a shorter amount of time, but will require more trips to the DB.
Zero or below will turn off batching and system will attempt to update everything in one go.

Intended for play and match per size/type/traffic of DB and system.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/command/DatabaseSettingsCommand.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/command/DatabaseSettingsCommand.java
@@ -15,6 +15,7 @@ public class DatabaseSettingsCommand {
     private String JNDIName;
     private int mysqlVarcharMaxlength;
     private String usertableQuote;
+    private int rowUpdateLimit;
 
     public DataSourceConfigType getConfigType() {
         return configType;
@@ -78,5 +79,13 @@ public class DatabaseSettingsCommand {
 
     public void setUsertableQuote(String usertableQuote) {
         this.usertableQuote = usertableQuote;
+    }
+
+    public int getRowUpdateLimit() {
+        return rowUpdateLimit;
+    }
+
+    public void setRowUpdateLimit(int rowUpdateLimit) {
+        this.rowUpdateLimit = rowUpdateLimit;
     }
 }
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/DatabaseSettingsController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/DatabaseSettingsController.java
@@ -56,6 +56,7 @@ public class DatabaseSettingsController {
         command.setJNDIName(settingsService.getDatabaseConfigJNDIName());
         command.setMysqlVarcharMaxlength(settingsService.getDatabaseMysqlVarcharMaxlength());
         command.setUsertableQuote(settingsService.getDatabaseUsertableQuote());
+        command.setRowUpdateLimit(settingsService.getDatabaseUpdateRowLimit());
         model.addAttribute("command", command);
     }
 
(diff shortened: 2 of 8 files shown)


randomnicode:
The ui is also part of this PR. It's the second commit (https://github.com/airsonic-advanced/airsonic-advanced/pull/31/commits/a6065d456cb1f1f8a4fdd8b2bd8b0b05f84189e5)

eharris:
Oops, overlooked that. Withdrawn. :-)

eharris:
Now I know why I was thinking it hadn't been done, I didn't see strings added for all the internationalization files.  Can you add those?

randomnicode:
I can, but I left it with just English because I thought that's the fallback. Correct me if I'm wrong here, but my thinking was that if a particular i8n var isn't found, it defaults to the one in the default locale properties file (`en`, in this case), is that right?

If that is the case, given I don't know the translation in most of the other languages, what would be the point of adding to those other files there? What would I add? If I just add English again, that seems redundant given they'd redirect to `en` anyway? A proper i18n translator would do a much better job, no?

eharris:
That's just been the convention, as it makes it clear that there's something that needs translation.  Without the entries there, someone who wants to do a translation isn't going to start from scratch with the english translation, they are only going to do the strings that aren't yet translated in the particular language file they know.  Without, someone would need to use some sort of tool that knows to go look in another file for missing entries, and I don't know what tools people might use.

But thinking a bit more about it, you might be right, as without looking at all the strings (even the ones that are already translated) side by side with the "source" strings, how would someone know if a change has been made to the source (English) that isn't reflected in the translation and needs to be updated?

I guess it all falls back to convention, and the convention in the original fork has been to update all the string files, even if a translation isn't yet available.

randomnicode:
Hmmm, I think we should modify the convention on account of the following:
- I think polluting an already translated file with untranslated strings renders a hodge podge of a file, with some stuff translated and some stuff not. I'd rather not pollute an existing completely-translated file. The default locale file should be used as a guide when someone is actually updating the translations, I assume that's what they do anyway.
- It keeps the burden of a PR light. If every time someone tries to update the UI or add a feature in the UI and has to update 30+ files for it with the same string, that's annoying, and it adds up. If I try to add 3 additions to the UI in 3 separate PRs, I'd have to modify 90+ files. That seems like undue burden for a PR writer, and that tiny bit of friction per addition adds to that hesitation of updating the UI. If I'm in a hurry, I'd rather just not deal with the UI portion than deal with updating 30 files for a tiny tiny change. I think the more frictionless we make it, the more people are amenable to making a desired change because the cost of development is much lowered for them. I know, for myself, initially I said I wasn't going to deal with the UI for this reason (I said I'd just make it an env var and deal with the UI at some other time)
- Translation seems to be an orthogonal concern to me, and I think should be handled separately. If a particular feature is interesting, I see no reason why a writer should be dealing with unrelated language files. I'm sympathetic to the plight of translation PRs, but we're adding a whole lot of extra process (update 30+ files for every UI change) for a minor and seldom used benefit (update a translation file)

Obviously all this is moot if it turns out that an i18n var cannot do a fallback to the default locale (in that case we should implement a graceful fallback implementation of the message source)

eharris:
That makes sense.  In that case, I'm fine with your approach as long as the fallback mechanism is working.  I haven't tested that myself though.

randomnicode:
Yeah, I think it'd be wise to test before we merge

PR #32, opened 2019-11-30
Add last scanned indices for tables that use them

randomnicode:
Several queries for artists, albums and media files are using last_scanned to query rows particularly when marking nonpresent entities or when expunging. This index should technically help for bigger dbs so the index is a factor in the query planner's arsenal.

### DIFF
--- a/airsonic-main/src/main/resources/liquibase/10.6/add-lastscanned-indices.xml
+++ b/airsonic-main/src/main/resources/liquibase/10.6/add-lastscanned-indices.xml
@@ -1,0 +1,35 @@
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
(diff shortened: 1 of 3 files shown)


randomnicode:
I think the performance concerns are certainly valid, and it's why I haven't merged this yet, but without any supporting data or benchmarks, I don't think either of us can say properly.

While the maintenance of an index during entry creation and update is bound to slow down stuff, finding an entry to delete will be faster, especially on a larger table, which is bound to be slower for a full table scans (which happens for each entry that needs to be deleted).

Let's say I need to delete 3 entries and add/update 3 entries (separately). Currently we'll do 3 table scans for each delete, and do a simple insert/update for 3. After the application of the index, we pay a tiny penalty for each insert/update, but in return we're saved from 3 whole table scans.

Worth it? I don't know. If the table isn't big, the query planner will likely not even use the index and opt for just the table scan, which means we're paying the penalty for no reason. If it's pretty big, then yeah, I think it would be, especially if you delete frequently.

I think an argument could be made that folks more likely add music more than delete. There's certainly some validity to that.

I could go either way. I think I'll at least keep the PR and not merge it unless it's needed.

eharris:
I didn't realize that the updates to the last-scanned value were being done individually, I had assumed they were done all at once (or batches).  If they aren't being done together, then I guess I don't understand why the query optimizer would ever consider using that index, as it has to be keying off something else unique anyway to get the right entry to update.

And if these updates to last-scanned are being done a single row at a time, that seems like something that needs to be fixed.

PR #53, opened 2019-12-29
Seekable transcodes

randomnicode:
Same PR as https://github.com/airsonic/airsonic/pull/1123

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/StreamController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/StreamController.java
@@ -255,6 +255,12 @@ public class StreamController {
                                 sendDummyDelayed(buf, out);
                             }
                         } else {
+                            if (fileLengthExpected != null && bytesWritten <= fileLengthExpected
+                                && bytesWritten + n > fileLengthExpected) {
+                                LOG.warn("Stream output exceeded expected length of {}. It is likely that "
+                                    + "the transcoder is not adhering to the bitrate limit or the media "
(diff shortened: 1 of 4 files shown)


eharris:
Ah thanks for pulling this in.  I was just about to work on submitting it but saw you had already done it.

airsonic-main/src/main/java/org/airsonic/player/controller/StreamController.java
```diff
@@ -230,25 +232,37 @@ public void handleRequest(HttpServletRequest request, HttpServletResponse respon
...
                         if (n == -1) {
                             if (isPodcast || isSingleFile) {
+                                // Pad the output if needed to avoid content length errors on transcodes
+                                if (fileLengthExpected != null && bytesWritten < fileLengthExpected) {
+                                    sendDummy(buf, out, fileLengthExpected - bytesWritten);
+                                }
                                 break;
                             } else {
-                                sendDummy(buf, out);
+                                sendDummyDelayed(buf, out);
```
randomnicode: out of curiosity, why is this sent after a delay?

eharris: I just preserved the original behavior, renaming the function so its name was a better indication of what it does.  I surmise it is attempting to keep the stream alive while waiting for some other event to occur, but I'm not really sure under what circumstances it was used.

randomnicode: okay, i'll try to dig in what's happening

PR #63, opened 2020-01-30
Secure creds

randomnicode:
Rehaul of the current login and credentials storage system to make them more secure. The existing credential system is not secure at all (stores everything in plain), and does not have flexibility to support multiple types of encodings. This causes multiple issues summarized in https://github.com/airsonic/airsonic/issues/69, summarily forcing the users to choose either between security or backwards-compatibility.

This PR centralizes a place to store credentials (external or airsonic), and provides several encoding schemes for users to choose. Users can be secure as well as backwards-compatible.

Two types of encoders are introduced:
- Nondecodable: These encode passwords in a form where the password itself isn't stored and isn't recoverable. You may only check if a given password is equal to the encoding, but you can never reverse the password from the encoding itself. This is more secure (since password isn't stored), but does not support the `t+s` form of authentication supported by Airsonic. The default encoder for this type is `bcrypt`. This type of encoder is new to Airsonic
- Decodable: These encode passwords in a reversible manner. The passwords may be recovered. This is less secure, but supports `t+s` style of Airsonic auth. Also credentials for third parties must be stored in this form. The default encoder for this type is `encrypt-aes-gcm` (which uses a shared key to encrypt data). Previously `plaintext` and `hex` were used in Airsonic that belong to this category.

At first migration, credentials are marked legacy and transferred to new tables. Old tables are left intact (and there is an admin action exposed to purge them, which would obviously not be backwards compatible, but more secure). Users may either change their credentials to a nondecodable format (more secure), or can confirm they want to retain decodable formats (less secure, more compatible). Additionally admins can migrate the whole system too. If no decision is made, the credentials remain marked as legacy and future versions of Airsonic will force-migrate everything for security (bcrypt for airsonic creds, encrypted for third-party).

The ability to add multiple credentials for the same account is also provided. In fact, credentials once entered may be considered immutable. The only way to change credentials is to add new ones (and remove old ones if needed). Adding multiple credentials for Airsonic allows Airsonic to be logged in with different passwords for the same account. However, multiple credentials for third-parties only uses the latest (most recently updated) credential when trying to log in. This can be changed in future PRs.

A new UI interface provides administration, the older Password Settings has been removed.

Salient features:
- Adds a new table called user_credentials, which supports multiple encoding schemes
- Migrates credentials data from other tables. Format is detected and stored (hex or opentext)
  - airsonic credentials from users.password
  - listenbrainz credentials from user_settings.listenbrainztoken
  - last.fm credentials from user_settings.lastfm_password and user_settings.lastfm_username
- Provides capability to change encoding schemes for given creds (if possible) on the fly
- Uses bcrypt as the default scheme for nondecodable encoders and encrypt-aes-gcm for decodable
  - Can change default schemes if wanted for some reason
- Ability to add multiple creds per profile (say one for secure password and another for backwards-compatible password)
- Addition of encryption as an encoder (default for decodable passwords)
- Admin capabilities to migrate the entire system
- Rewritten REST security filter to use the new auth provider
- New auth provider that iterates through and tries to match all the stored credentials until one is found that matches.

UI snaps:

image

image


randomnicode:
@eharris @fxthomas I'm looking for feedback particularly regarding terminology and names of columns.
There is some inconsistency in usage of terms, and I'd like to come to a consensus.
I started out initially using `location` (where credentials apply to, like last.fm or airsonic) and `type` (what we use to encode them, like bcrypt or plaintext), but then switched to `app` and `encoder` respectively, when I got to the UI part. Now I seem to use them interchangeably, and I have second thoughts about that. Any opinions or votes what sounds clearer? Any better options to name them? I'm leaning towards `encoder` instead of `type`, and slightly preferential to `app` instead of location, but I don't really have much of a preference.

Any thoughts about UI layout? Not the best but I thought future PRs could always address it better, but if there is a suggestion, I can add it.

fxthomas:
`app` sounds okay, but why not `type` or `encoding`? The DB will not expose encoders directly, it will expose types of encoding methods (which are coincidentally implemented as encoders, but this could be changed in principle).

Why not rely on Spring's mechanism for selecting encoders/decoders (`{<encoder>}<encoded password>`) and build our own?

randomnicode:
Apologies for the miscommunication. Let me try and explain it better:
- `app` (used in UI) or `location` (used in DB column) refers to what the creds are for.
  - Possible values: airsonic, last.fm, listenbrainz. More if we add more external parties (like some lyrics services, other scrobblers etc)
  - Need to settle on ONE of the following so we may use it consistently throughout. Currently this PR uses these two terms interchangeably, and I think that'll be confusing
    - `app`
    - `location` 
    - some other name
- `encoder` (used in UI) or `type` (used in the DB) refers to how they're encoded.
  - Possible values: noop, hex, bcrypt
  - Need to settle on ONE of the following so we may use it consistently throughout, instead of different terms in the DB and different in the UI
    - `type`
    - `encoder`
    - `encoding`
    - some other name

> Why not rely on Spring's mechanism for selecting encoders/decoders ({<encoder>}<encoded password>) and build our own?

The format is only used by the DelegatingPasswordEncoder to select the appropriate decoder for matching.

We _do_ use the format (we construct it) when we use the DelegatingPasswordEncoder to match: https://github.com/airsonic-advanced/airsonic-advanced/pull/63/files#diff-b209f1eb72209c4975fcc15907bd8874R48

But sometimes we need to use the base encoder directly without dealing with the DelegatingPasswordEncoder. In that case the format isn't appropriate. For instance, passing that format directly to the bcrypt encoder will result in an error (it already knows it's bcrypt, "{bcrypt}" prefix isn't needed).

Thus, we need to keep the encoder separate because occasionally we need to decode, or reencode (like legacy migrations) without involving the DelegatingPasswordEncoder. Also keeps it cleaner. There is clear delineation in case we ever want to use something other than spring.

randomnicode:
Alright, 23 days seems long enough, and I don't want it to keep lingering.

This is getting merged. Any issues we can resolve in follow up PRs

PR #123, opened 2020-03-09
Trim user_credentials encoder

randomnicode:
HSQLDB 2.5 for some unknown reason puts a space after 'legacyhex' during automigrations.

A fix is to just trim the result of the migration


randomnicode:
Turns out that HSQLDB 2.5 is treating string literals in CASE WHEN statements as chars and thus padding with blank space. This may be resolved by setting a property on the db, therefore this PR isn't necessary

PR #130, opened 2020-03-15
DataTables for optimized rendering of tables

randomnicode:
This PR follows up from #129 and actually starts using a library (DataTables) for faster rendering of tablular data. It also allows additional advantages like paging, reordering, highlighting, searching etc.

This PR only introduces changes to PlayQueue handling of tables as that seemed more urgent given the rendering slowness described in #129 

It also introduces the following stylistically:
- hover highlighting
- currently playing highlighting
- fixes playqueue height issues
- removes animation of playqueue completely
- fixes dark themes using light mediaelement css
- css refactors


khannurien:
Hi! Thank you for the hard work.

Would it be possible to add a user preference for the default number of rows in tables? I would like to default to "All" instead of 10.

randomnicode:
Sure. Added as issue #154  to track it

Issue #154, opened 2020-03-27
Default number of rows in tables


randomnicode:
Hi! Thank you for the hard work.

Would it be possible to add a user preference for the default number of rows in tables? I would like to default to "All" instead of 10.

_Originally posted by @khannurien in https://github.com/airsonic-advanced/airsonic-advanced/pull/130#issuecomment-604473536_

randomnicode:
These are variables to be considered:
- User-specified rows for which tables
  - One per table (that's at least 3 tables, and more if other tables crop up)
  - Only specifiable for Playqueue, everything else gets a default number
- Performance
  - Large number of rows will slow down page performance
    - Will need to add the Scroller extension to keep performance
      - Will need to look at the row reordering/click handling

khannurien:
As I'm heavily using playlists, I'm particularly interested in begin able to default to "All" for these.

Regarding "convention over configuration" -- I think the album view of all the views could be defaulted to "All" without user setting. It feels silly to have to use page 2 for one or two tracks.

randomnicode:
Merged, @khannurien please test the merged release and see if this issue can be closed?

randomnicode:
> I think the album view of all the views could be defaulted to "All" without user setting

Doesn't work for people working with massive directories where they dump all songs in one folder. There will be significant performance drawbacks for trying to draw that kind of large table.

Just for usecase profiling, why is it that you need to default to All? What is it that paging hinders in your normal workflow? The data is all client side so the lag time between drawing pages should be minimal and will save you some performance implications that you might get for a large table.

khannurien:
Your patch seems to work flawlessly. Thank you!

> Just for usecase profiling, why is it that you need to default to All?

Albums usually don't exceed 20 or 30 tracks, and I'm used to being able to see every track in a playlist when I'm working on it.
I guess these are just habits, and I'm glad I'm now able to stick to it.

> Doesn't work for people working with massive directories where they dump all songs in one folder.

Fair enough, that's why I advocated for a user setting in the first place! :-)

randomnicode:
Sorry can you please point out what the bug is?

khannurien:
Sorry I haven't been clear enough -- it is the padding of the top bar:

![image](https://user-images.githubusercontent.com/31770422/78803782-b2944f80-79bf-11ea-9519-9e6af7498bee.png)

It used to be narrower, like that:

[...]

PR #182, opened 2020-04-08
Set top panel iframe height via javascript

randomnicode:
The top panel remains an iframe (hasn't been converted yet).

This poses layout problems, especially since it is the first element in the layout. Because the iframe contents haven't loaded yet, the iframe height is set by the useragent to be 150px. Since that is the initial basis for flex, it goes from there and stays at that height.

The approach here is to wait for the contents of the iframe to load and then set the iframe height forcibly afterwards via javascript. This means we can take it out of the flex equation altogether, so the bottom panel effectively flexes to take the rest of the space.

### DIFF
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/index.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/index.jsp
@@ -16,11 +16,9 @@
         }
         .main-navigation {
             width: 100%;
-            height: 0%;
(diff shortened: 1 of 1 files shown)


randomnicode:
Fixes issue pointed in https://github.com/airsonic-advanced/airsonic-advanced/issues/154#issuecomment-611032179

khannurien:
That was fast! Well done.

PR #210, opened 2020-04-17
Java 11

randomnicode:


randomnicode:
tests pass, lgtm just seems to be having a fit likely because it wasn't notified about the language change

Issue #202, opened 2020-04-16
Stuck on first scan while first start after upgrade to Airsonic-Advanced


Faeb35:
## Problem description

I just updated from Airsonic to Aisonic-Advanced. I removed every file except the internal database. When starting Airsonic-Advanced for the first time it seams to do a complete full rescan of my medias (which takes very long for my 112000 mp3 files). When updating the genres (according to the log) it stuck for about 25 minutes and cosuming the power of one cpu (25% CPU, I have 4 cores). During this 25 minutes I was not able to access the Airsonic website. I always got a timeout when loading the page. Before updating the genres and while scanning the medias I was able to access the website without any issue.
After doing this whole thing several times and finally waiting patiently enough it came over this and finished this step. I think updating the genre isn't running on a own thread.

### Steps to reproduce

1. Take a copy of a big airsonic database.
2. Start airsonic-advanced for the first time.
3. Wait until media scan has finished and genres are updating

## System information

 * **Airsonic version**: Edge Release 10.6.0-SNAPSHOT.20200414043004
 * **Operating system**: Windows Server 2012
 * **Java version**: Java(TM) SE Runtime Environment (build 1.8.0_241-b07)
 * **Proxy server**: IIS
 * **Client**: Tested in Firefox
 * **Language**: English

## Additional notes

2020-04-16 21:14:38.406  INFO --- o.a.p.s.MediaScannerService              : Scanned media library with 115751 entries.
2020-04-16 21:14:46.213  INFO --- o.a.p.s.MediaScannerService              : Scanned media library with 115855 entries.
2020-04-16 21:14:46.230  INFO --- o.a.p.s.MediaScannerService              : Persisting albums
2020-04-16 21:14:46.291  INFO --- o.a.p.s.MediaScannerService              : Persisting artists
2020-04-16 21:14:46.353  INFO --- o.a.p.s.MediaScannerService              : Marking present files
2020-04-16 21:14:46.353  INFO --- o.a.p.s.MediaScannerService              : Persisting genres
2020-04-16 21:**14**:47.931  INFO --- o.a.p.s.MediaScannerService              : Updating genres
2020-04-16 21:**40**:31.175  INFO --- o.a.p.s.MediaScannerService              : Marking non-present files.
2020-04-16 21:40:31.207  INFO --- o.a.p.s.MediaScannerService              : File marking complete
[...]

PR #240, opened 2020-04-22
Update in batches because Postgres wire cannot handle large number of params

randomnicode:
Postgres driver can only send a certain number of params over the wire in one go (32767).
This poses issues for `IN` statements with a large number of params (such as a large number of media files being marked present simultaneously).
Base airsonic updated files one-by-one and thus does not come across this issue.

This fix introduces batchedUpdates so updates are still sent en masse but in batched statement form in batches that are below the postgres limit (even if the db isn't postgres). Those drivers that do not support batched updates (maybe HSQLDB 1.8?) will send statements over one by one.

The overall outcome is that the statement won't yield exceptions for postgres for a large number of params, BUT it will most certainly be slower than a large IN query because the query planner won't be able to optimize for all the args together, instead will see each one by one. The degree of slowness is also additionally compounded if the driver does not support batch updates so the updates go over the wire one by one.

An additional improvement may be to break up chunks of args and still send them in IN statements, but the degree of improvement needs to be measured. This was implemented in the second commit.

For more info, a sample issue is here:
https://github.com/jOOQ/jOOQ/issues/5701

The exception is seen in issue #202 and this PR should fix it.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/dao/AbstractDao.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/dao/AbstractDao.java
@@ -109,13 +109,13 @@ public class AbstractDao {
         int batchSize = 30000 / batchArgs.stream().findAny().map(x -> x.length).orElse(1);
         LOG.trace("Executing query: [{}]", sql);
         int[][] result = getJdbcTemplate().batchUpdate(sql,
-                batchArgs.parallelStream().map(AbstractDao::convertToDBTypes).collect(Collectors.toList()),
-                batchSize,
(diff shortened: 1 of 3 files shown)


Faeb35:
Hi randomnicode,
I just ran my test again an let my media library scan with using PostgreSQL like described in [#202](https://github.com/airsonic-advanced/airsonic-advanced/issues/202).
I can totaly confirm, that this pull request fixed the issue and scanning the media library now workes like a charm. 
Awesome work!
Thank you very much for all your hard work!!

Issue #224, opened 2020-04-20
"Show X entries" not saved and in wrong order


Faeb35:
## Problem description

Like in the playing queue or when opening an album you get the following option:
![grafik](https://user-images.githubusercontent.com/9485516/79765771-56b6b880-8327-11ea-90b3-6307b8223f5f.png)
Two things with this drop down:
1. Is there a place, where I can set a default value for that? Or should this be saved automatically when changing the value? For me it always goes back to 40.
2. 40 is on the bottom instead after 20 and before 50.

## System information

 * **Airsonic version**: Edge Release 10.6.0-SNAPSHOT.20200414043004
 * **Operating system**: Windows Server 2012
 * **Java version**: Java(TM) SE Runtime Environment (build 1.8.0_241-b07)
 * **Proxy server**: IIS
 * **Client**: Tested in Firefox
 * **Language**: English

randomnicode:
`40` is a custom non-standard option and therefore appears separately and not in order. For some reason earlier versions of airsonic used 40 as the default before and so you're now seeing it.

Newer users I think have one of the standard pagination options autoselected (put in airsonic-advanced).

Yes, it can be customized here: Settings -> Personal -> Pagination size for tables (-1 shows everything)

randomnicode:
Close this issue please?

Faeb35:
Hi randomnicode
Thank you very much for the explanation and your time!
Now I remember, I was the person who set the value of 40 there! What a shame :-(
Very very sorry for opening this unecessary issue!

awsms:
> Yes, it can be customized here: Settings -> Personal -> Pagination size for tables (-1 shows everything)

 If I set it to -1, 249 or even 294502592, it will still show 7 elements on a page. I'm using the latest stable release 10.6.0

randomnicode:
I think you're looking at the subdirectories table. That is hardcoded to 7.
https://github.com/airsonic-advanced/airsonic-advanced/blob/416e772afd7ba62d118d3b9edf171c5f27d89c20/airsonic-main/src/main/webapp/WEB-INF/jsp/mediaMain.jsp#L438-L439

The regular files table is populated with the user preference

PR #274, opened 2020-04-30
Change pagination setting to dropdown menu

Faeb35:
This is my first pull request on Airsonic ever (and the third ever), so don't judge me if there are any mistakes ;-)

I changed the free text input field, where you can set the "numbers of rows displayed per page", to a dropdown field and moved it on the top of the personal settings, where the other dropdowns are.
As this list contains the word "All" I added it to the messages and translated it to German (the other languages I don't speak unfortunately). Additionally I translated the "All" from the dropdown in the mediaMain table as well by using messages.

I personally already had an "issue" with that setting (#224) and was thinking this would be a very easy "low haning fruit" for a first pull request. I'm hoping that this PR gives a tiny little bit of extra usability.


Faeb35:
I just realized, that there were some unintended changes in ResourceBundle_pt_BR.properties!
Does anybody know how this happend? and how this can be fixed? Sorry for that!

Faeb35:
How can the "Show XXX Entries" be translated?

airsonic-main/src/main/resources/org/airsonic/player/i18n/ResourceBundle_pt_BR.properties
```diff
@@ -137,7 +138,7 @@ search.index=O índice de pesquisa está sendo criado. Tente mais tarde.
 search.hits.none=Nada encontrado.
 search.hits.more=Mais
 search.hits.artists=Artistas
-search.hits.albums=Álbuns
+search.hits.albums=�?lbuns
```
randomnicode, line 141: Encoding issues?

randomnicode:
A few comments on this PR:

- This setting was actually kept as an input box for a reason. A dropdown takes away the feature of being able to set a _custom_ pagination size. With this PR you can only select among a certain (predefined) options. If a particular user wants to see `23` lines per table (due to screen constraints or whatever), s/he won't be able to do that. It was much simpler to do a dropdown in the original PR, but was opted against (see the code above the pagination size settings, which allows the setting of a custom size).
- I think there is some issue with your text editor encoding. It is converting some characters it shouldn't be
- I think the initiative to i18n the "All" dropdown is good, but there is already an "All" in the resource bundle under `home.alphabetical.title=All`. I think we should consolidate. Maybe rename `home.alphabetical.title` to `common.all` apply that everywhere?
- Either way, the i18n stuff should be a separate PR where the sole focus is just i18n

randomnicode:
> How can the "Show XXX Entries" be translated?

Datatables is the client-side library. It has its own internationalization scheme that you can try bringing in.

Faeb35:
Hi @randomnicode,

thanks for your feedback! My first pull request doesn't seems to be too "successfull" ;-)

I truely understand your point of giving the possibility to choose a custom value. That totaly make sense to me. I didn't think about this more deeply, like that users would like to choose a custom value e.g. depending on the screen size. For the screen size, I think it even might be hard to find a value which suites everywhere. This table is used e.g. in the player queue, playlists, album view, and Airsonic is used probably also with different screens with different screen resolutions (having two screens on the computer or use Airsonic on multiple computer).

I was more thinking about this more from a usability point of view, how users are used to use controlls like this. 
- All applications I now are working like this, they just provide you a set of values for the pagination setting. So users are used to use it like that. If it's different, then users getting confused. Thats what happend to me.
- It would make it more recognizable/consistant to have the same dropdown in the settings and in the table view.
- No use of the more "technical" -1

I defenitly will/would (if I don't cancle this PR) remove all the translation related stuff from this PR.

What options do we have:
a) Add more options to the dropdown
b) Add a checkbox next to the dropdown with thext "Use custom value". When you select this value, the dropdown is getting replaced with an text field (like the one was there before). Eigther the dropdown or the text field could be display: none.
c) Cancel this PR

grafik

Option b) is my favorit ;-)

Faeb35:
I just realized, that there's currently also no check about what you enter in the current "Pagination size for tables" field. If you enter text, it leads to an 400 Bad Request error. So I think this could be addressed as well.

Faeb35:
Just pushed a new version. Should hopefully satisfy all needs.

grafik

zrav:
Any progress with getting this merged?

randomnicode:
This is a little out of date since there are now multiple settings for different tables. As it stands, this cannot be merged

Issue #201, opened 2020-04-16
Getting horizontal scrollbar when opening player queue


Faeb35:
## Problem description

I'm using Firefox with the "Tree Style Tab" extension. This extension shows me all the opened tabs on the left side of the browser. So my websites are not having the fill with.
I just started to try airsonic-advanced and realized, that I get a horizontal scroll bar, when I open the player queue. When I close the sidebar of the browser extension all is working find.

### Steps to reproduce

1. Install "Tree Style Tab" extension.
2. Open player queue
3. You'll see the scrollbar

Edit:
The issue seems to go away, when I zoom out with the browser and then zoom in again.

Faeb35:
I noticed this issue as wall without the Tree Style Tab extension. I only tested in Firefox. As my screen on which I use Airsonic Advanced isn't to wide, I usually zoom out to 90%. The issue seems to be related to the zoom factor used by the browser...

Faeb35:
I just did some troubleshooting. Conclusion: If I remove the width css attribute (or set it to 100%) on the following to nodes, the issue is fixed:
div.dataTables_scrollHeadInner
and
div.dataTables_scrollHeadInner > table

When zooming in and out (or changing the windows size) this width css seems to get recalculated, an set again if changed manually.

Seems to come from datatables.js line 5112 or jquery.dataTables.js line 5100, from "scroll.sXInner". As datatables.js looks like an external plugin/library it probably doesn't make any sense to do changes there I think.

Faeb35:
This code would resolve the issue, but it's more a workaround then a propper solution...

```
$(window).on( "resize", function (e) {
                var scrollHead = $("#playQueueMusic_wrapper > .dataTables_scroll > .dataTables_scrollHead > .dataTables_scrollHeadInner");
                if (scrollHead.length > 0) {
                    setTimeout(function() {
                        $(scrollHead).css("width", "100%");
                        $(scrollHead).find("table.dataTable").css("width", "100%");
                    }, 0);
                }
            } );
```

PR #278, opened 2020-05-01
Remove selections after action in play queue

Faeb35:
When I selected some tracks in the play queue and used the action "Add to playlist" the tracks were still selected after this action had been finished. Usually I'll add some more tracks to a playlist later, so I always had to remove the selection manually after adding them to the playlist. If I forgot this I ended up having some tracks twice (or more) in my playlist.

I think the process is as follow:
1. User likes to to an action with his tracks
2. User selects some tracks
3. User does the action
4. As the action is finished now, he won't need the selection anymore

So I made this pull request to remove selection, after the action in the play queue has been finished. To make it consistent I also removed the selection after downloading. For removal it wasn't needed, as there's nothing to deselect anymore...

### DIFF
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
@@ -891,6 +891,7 @@
                 // change page
                 top.main.location.href = "playlist.view?id=" + playlistId;
             }
+            this.selectAll(false);
             $().toastmessage("showSuccessToast", toastMsg);
         },
 
@@ -1067,6 +1068,7 @@
                 this.onRemoveSelected();
             } else if (id == "download" && selectedIndexes != "") {
                 location.href = "download.view?player=" + this.player.id + "&" + selectedIndexes;
+                this.selectAll(false);
             } else if (id == "appendPlaylist" && selectedIndexes != "") {
                 this.onAppendPlaylist();
             }


Faeb35:
@randomnicode What do you think about this? 
If it's fine with you, I would start implementing the same in the album view to get it consistant.

randomnicode:
So my personal use case is to add a bunch of tracks to multiple playlists. For that purpose, it's better to keep it selected:
-  Select a bunch of tracks
- Perform action 1 (add to Playlist 1)
- Now they're still selected, so easily able to perform action 2 (add to Playlist 2)
... repeat

From a beyond personal, higher perspective, my view is the following:
- If we unselect after every operation, it disallows us from chaining operations on the same group of tracks. Each track now needs to be reselected.
- It's much more difficult to reselect a bunch of tracks (sometimes strewn across a bunch of pages) than it is to unselect them once (clear between operations)

So I'd prefer not unselecting them after each operation because it allows me to do a bunch of operations on the same group of files.

You might look into a "select all" checkbox above the table rows?

Faeb35:
Thanks for your reply!
I totally see your point. Not all users are using Airsonic the same way, and I also agree that reselecting the tracks again would be a bigger hustle, then what's my use case.

Instead of having a "Select all" I rather would like to have a "Select none" button ;-)
But this one is already there in the "More actions...".

Before submiting this PR I was thinking about placing a "Select none" button somewhere, e.g. next to "Now playing" would be easy to achieve, but as this is alread under "More actions..." I thought it wouldn't be smart to have it twice...

Faeb35:
I was also thinking about placing a little "X" just next to "... rows selected".
grafik

But for that I think a would need to change the datatables.js library. But I think this wouldn't be clever. I was questioning this myself as well in #201. 
Other option would be to create a little jQuery maniplulation the datatables.js DOM after rendering and adding the X. But I'm not a fan of doing that... 

@randomnicode What's your opinion about changing things in datatables.js? Do it, and have to struggle with it after comming updates on datatables.js? Or doing jQuery manipulation after rendering? Are there other approaches?

PR #285, opened 2020-05-07
Fix download issue of files in album view

Faeb35:
Under Windows 10 with OpenJDK 14, I was not able to do downloads in the view where I browse an album (only download from player queue was working). There was always a 500 error with NullPointer Exception when starting the download. When selecting multiple tracks for download, there was always an additional path which was null.
To filter that I add
`.filter(p -> p != null)`

When downloading a single file it always created a zip file instead of just downloading the file. For that I had to add the following code, additionalFiles had a length of 1 but this element was null.
` || (additionalFiles.length == 1 && additionalFiles[0] == null)`

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/DownloadController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/DownloadController.java
@@ -205,7 +205,7 @@ public class DownloadController {
                             settingsService.getDownloadBitrateLimiter(),
                             statusSupplier,
                             statusCloser,
-                            (input, status) -> {}),
+                        (input, status) -> {}),
                     path.getFileName().toString(),
                     file.getFileSize(),
                     changed);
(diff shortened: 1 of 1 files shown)


Faeb35:
Not really sure, why getting the Maven errors 

```
[ERROR] /home/runner/work/airsonic-advanced/airsonic-advanced/airsonic-main/src/main/java/org/airsonic/player/controller/DownloadController.java:208: 'lambda arguments' has incorrect indentation level 28, expected level should be 24. [Indentation]
Audit done.
...
[ERROR] src/main/java/org/airsonic/player/controller/DownloadController.java:[208] (indentation) Indentation: 'lambda arguments' has incorrect indentation level 28, expected level should be 24.
```
Don't really see a difference there to the code before, except formating...

randomnicode:
How about instead of letting null pollute the method and then checking for it everywhere, we just change the method signature to take in a collection instead of a varargs param?

It seems that the primary culprit that's sending nulls is sitting here:
https://github.com/airsonic-advanced/airsonic-advanced/pull/285/files#diff-e3df54a4b41d653992a37f5793fd29a9R136

so how about we change the last arg here:
https://github.com/airsonic-advanced/airsonic-advanced/pull/285/files#diff-e3df54a4b41d653992a37f5793fd29a9R184
from `Path...` to a `Collection<Path>`?

Then all the methods calling this method explicitly can send either an empty collection or a boxed collection of Paths?

You can retain the safety nullcheck in the stream (the concat bit will need to change too)

PR #297, opened 2020-05-11
Fix album download NPE

randomnicode:
Similar to and supersedes #285

Thanks to @Faeb35


Faeb35:
Thank you for taking over! I was a bit short on time in the last days...

PR #301, opened 2020-05-14
Podcast episodes more actions dropdown

Faeb35:
grafik

This pull request adds a "More actions..." dropdown to the Podcast Episodes view. It has the same functionalities as the dropdowns in the player queue and the album view. Only the "Share" option is missing. This one was not straight forward to implement. The issue here is, that the external player doesn't have access to the podcast folder.

For that pull request I reused some existing websocket functionalities:
- To get index of the media file in the media folder (needed for the download)
- Playlist handling

If you select some episodes, which are not downloaded yet, you'll get an error toast message for the selected episodes. This text would need to be translated:
grafik

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/DownloadController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/DownloadController.java
@@ -103,11 +103,11 @@ public class DownloadController {
 
     @GetMapping
     public ResponseEntity<Resource> handleRequest(Principal p,
-                                                  @RequestParam Optional<Integer> id,
-                                                  @RequestParam(required = false) Integer playlist,
-                                                  @RequestParam(required = false) Integer player,
-                                                  @RequestParam(required = false, name = "i") List<Integer> indices,
-                                                  ServletWebRequest swr) throws Exception {
+            @RequestParam Optional<Integer> id,
+            @RequestParam(required = false) Integer playlist,
+            @RequestParam(required = false) Integer player,
+            @RequestParam(required = false, name = "i") List<Integer> indices,
+            ServletWebRequest swr) throws Exception {
         User user = securityService.getUserByName(p.getName());
         Player transferPlayer = playerService.getPlayer(swr.getRequest(), swr.getResponse(), false, false);
         String defaultDownloadName = null;
@@ -165,8 +165,8 @@ public class DownloadController {
      */
     private static long computeCrc(Path file) throws IOException {
         try (InputStream is = Files.newInputStream(file);
-             BufferedInputStream bis = new BufferedInputStream(is);
-             CheckedInputStream cis = new CheckedInputStream(bis, new CRC32())) {
+            BufferedInputStream bis = new BufferedInputStream(is);
+            CheckedInputStream cis = new CheckedInputStream(bis, new CRC32())) {
             byte[] buf = new byte[8192];
             while ((cis.read(buf)) != -1) {
                 continue;
@@ -205,8 +205,7 @@ public class DownloadController {
                             settingsService.getDownloadBitrateLimiter(),
                             statusSupplier,
                             statusCloser,
-                            (input, status) -> {
-                            }),
+                            (input, status) -> {}),
                     path.getFileName().toString(),
                     file.getFileSize(),
                     changed);
@@ -242,7 +241,7 @@ public class DownloadController {
                 // start a new thread to feed data in
                 new Thread(() -> {
                     try (PipedOutputStream pout = new PipedOutputStream(pin);
-                         ZipOutputStream zout = new ZipOutputStream(pout)) {
+                        ZipOutputStream zout = new ZipOutputStream(pout)) {
                         zout.setMethod(ZipOutputStream.STORED); // No compression.
                         pathsToZip.stream().forEach(LambdaUtils.uncheckConsumer(f -> {
                             status.setFile(f.getKey());
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/DownloadController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/DownloadController.java
@@ -103,11 +103,11 @@ public class DownloadController {
 
     @GetMapping
     public ResponseEntity<Resource> handleRequest(Principal p,
-            @RequestParam Optional<Integer> id,
-            @RequestParam(required = false) Integer playlist,
-            @RequestParam(required = false) Integer player,
-            @RequestParam(required = false, name = "i") List<Integer> indices,
-            ServletWebRequest swr) throws Exception {
+                                                  @RequestParam Optional<Integer> id,
+                                                  @RequestParam(required = false) Integer playlist,
+                                                  @RequestParam(required = false) Integer player,
+                                                  @RequestParam(required = false, name = "i") List<Integer> indices,
+                                                  ServletWebRequest swr) throws Exception {
         User user = securityService.getUserByName(p.getName());
         Player transferPlayer = playerService.getPlayer(swr.getRequest(), swr.getResponse(), false, false);
         String defaultDownloadName = null;
@@ -165,8 +165,8 @@ public class DownloadController {
      */
     private static long computeCrc(Path file) throws IOException {
         try (InputStream is = Files.newInputStream(file);
-                BufferedInputStream bis = new BufferedInputStream(is);
-                CheckedInputStream cis = new CheckedInputStream(bis, new CRC32())) {
+             BufferedInputStream bis = new BufferedInputStream(is);
+             CheckedInputStream cis = new CheckedInputStream(bis, new CRC32())) {
             byte[] buf = new byte[8192];
             while ((cis.read(buf)) != -1) {
                 continue;
@@ -195,7 +195,7 @@ public class DownloadController {
         }
 
         if (indices.size() == 1 && (additionalFiles == null || additionalFiles.size() == 0)) {
-            // single file
+		// single file
             MediaFile file = files.get(indices.get(0));
             Path path = file.getFile();
             long changed = file.getChanged() == null ? -1 : file.getChanged().toEpochMilli();
@@ -205,7 +205,8 @@ public class DownloadController {
                             settingsService.getDownloadBitrateLimiter(),
                             statusSupplier,
                             statusCloser,
-                        (input, status) -> {}),
+                            (input, status) -> {
+                            }),
                     path.getFileName().toString(),
                     file.getFileSize(),
                     changed);
@@ -241,7 +242,7 @@ public class DownloadController {
                 // start a new thread to feed data in
                 new Thread(() -> {
                     try (PipedOutputStream pout = new PipedOutputStream(pin);
-                            ZipOutputStream zout = new ZipOutputStream(pout)) {
+                         ZipOutputStream zout = new ZipOutputStream(pout)) {
                         zout.setMethod(ZipOutputStream.STORED); // No compression.
                         pathsToZip.stream().forEach(LambdaUtils.uncheckConsumer(f -> {
                             status.setFile(f.getKey());
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/PodcastChannelController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/PodcastChannelController.java
@@ -53,13 +53,14 @@ public class PodcastChannelController {
 
         Map<String, Object> map = new HashMap<>();
         ModelAndView result = new ModelAndView();
-        result.addObject("model", map);
 
         int channelId = ServletRequestUtils.getRequiredIntParameter(request, "id");
 
         map.put("user", securityService.getCurrentUser(request));
         map.put("channel", podcastService.getChannel(channelId));
         map.put("episodes", podcastService.getEpisodes(channelId));
+
+        result.addObject("model", map);
         return result;
     }
 
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
@@ -27,9 +27,9 @@
     <script type="text/javascript" src="<c:url value='/script/utils.js'/>"></script>
     <script type="text/javascript" language="javascript">
         function init() {
-            top.StompClient.subscribe("pdcactChannel.jsp", {
+
+            top.StompClient.subscribe("podcactChannel.jsp", {
                 '/user/queue/playlists/writable': function(msg) {
-                    console.log('playlistSelectionCallback');
                     playlistSelectionCallback(JSON.parse(msg.body));
                 },
                 '/user/queue/playlists/files/append': function(msg) {
@@ -82,10 +82,7 @@
         function appendPlaylist(playlistId) {
             $("#dialog-select-playlist").dialog("close");
 
-            //var mediaFileIds = filesTable.rows({selected:true}).data().map(function(d) { return d.id; }).toArray();
             var mediaFileIds = getSelectedEpisodesMediaId();
-            console.log(mediaFileIds);
-
             top.StompClient.send("/app/playlists/files/append", JSON.stringify({id: playlistId, modifierIds: mediaFileIds}));
         }
 
@@ -208,6 +205,8 @@
     <c:forEach items="${model.episodes}" var="episode" varStatus="i">
 
         <tr>
+            <td class="fit"><input type="checkbox" id="episode${i.index}" value="${episode.id}" mediafileid="${episode.mediaFileId}" /></td>
+
             <c:choose>
                 <c:when test="${empty episode.mediaFileId or episode.status ne 'COMPLETED'}">
                     <td colspan="4"></td>
@@ -222,8 +221,6 @@
                     </c:import>
                 </c:otherwise>
             </c:choose>
-
-            <td class="fit"><input type="checkbox" id="episode${i.index}" value="${episode.id}" mediafileid="${episode.mediaFileId}" /></td>
 
             <td class="truncate">
                 <span title="${episode.title}" class="songTitle">${episode.title}</span>
@@ -267,7 +264,7 @@
 </table>
 
 <select id="moreActions" class="pagetype-dependent type-album type-video" onchange="actionSelected(this.options[selectedIndex].id);" style="margin-bottom:1.0em">
-    <option id="top" selected="selected"><fmt:message key="main.more.selection"/></option>
+    <option id="top" selected="selected"><fmt:message key="playlist.more"/></option>
     <option id="selectAll">&nbsp;&nbsp;<fmt:message key="playlist.more.selectall"/></option>
     <option id="selectNone">&nbsp;&nbsp;<fmt:message key="playlist.more.selectnone"/></option>
     <c:if test="${model.user.downloadRole}">
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
@@ -26,9 +26,20 @@
 
     <script type="text/javascript" src="<c:url value='/script/utils.js'/>"></script>
     <script type="text/javascript" language="javascript">
+        var podcastEpisodes = [];
         function init() {
+            <c:forEach items="${model.episodes}" var="episode" varStatus="i">
+            var episode = {};
+            episode.id = '${episode.id}';
+            episode.mediaFileId = '${episode.mediaFileId}';
+            episode.status = '${episode.status}';
+            podcastEpisodes.push(episode);
+            </c:forEach>
 
-            top.StompClient.subscribe("podcactChannel.jsp", {
+            top.StompClient.subscribe("podcastChannel.jsp", {
+                '/user/queue/mediafile/directory/get': function(msg) {
+                    getMediaDirectoryCallback(JSON.parse(msg.body));
+                },
                 '/user/queue/playlists/writable': function(msg) {
                     playlistSelectionCallback(JSON.parse(msg.body));
                 },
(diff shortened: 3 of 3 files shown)


airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
```diff
@@ -1,45 +1,148 @@
...
+        var podcastEpisodes = [];
         function init() {
-            $("#dialog-delete").dialog({resizable: false, height: 170, autoOpen: false,
+            <c:forEach items="${model.episodes}" var="episode" varStatus="i">
+            var episode = {};
+            episode.id = '${episode.id}';
+            episode.mediaFileId = '${episode.mediaFileId}';
+            episode.status = '${episode.status}';
+            podcastEpisodes.push(episode);
+            </c:forEach>
```
Faeb35, line 37: Not quite sure if this is the best way to pass variables from the server to client side ;-)
Is there a better way doing that (whitout using websockets)?

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
```diff
@@ -70,28 +179,46 @@
...
 
+        function getSelectedEpisodesMediaId() {
+            var result = [];
+            for (var i = 0; i < ${fn:length(model.episodes)}; i++) {
+                var checkbox = $("#episode" + i);
+                if (checkbox.is(":checked")) {
+                    if (podcastEpisodes[i].status == 'COMPLETED') {
+                        result.push(podcastEpisodes[i].mediaFileId);
+                    } else {
+                        $().toastmessage("showErrorToast", "Download of the episode needs to be completed for that action");
```
Faeb35, line 190: Is there a tool for doing the translation? What's the best method? 
Updating all the files manually sounds very exhausting.

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
```diff
@@ -70,28 +179,46 @@
...
+        <c:param name="coverArtSize" value="200"/>
+    </c:import>
 </div>
 
-<h1 id="name"><a href="podcastChannels.view"><fmt:message key="podcastreceiver.title"/></a> &raquo; ${fn:escapeXml(model.channel.title)}</h1>
+<h1 id="name"><a href="podcastChannels.view"><fmt:message key="podcastreceiver.title"/></a>
+    &raquo; ${fn:escapeXml(model.channel.title)}</h1>
 <h2>
-    <span class="header"><a href="javascript:top.playQueue.onPlayPodcastChannel(${model.channel.id})"><fmt:message key="common.play"/></a></span>
+    <span class="header"><a href="javascript:top.playQueue.onPlayPodcastChannel(${model.channel.id})"><fmt:message key="main.playall"/></a></span>
```
Faeb35, line 211: Updated this to show "Play all" instead of "Play", which is clearer and more consistent

randomnicode:
:eyes

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
randomnicode: This isn't websockets, this is Spring MVC (using JSP).

Here's the lifecycle of a page:
- Request comes in
- Goes to a controller
- Controller builds a _model_
- Controller forwards to a jsp template with that model (still within the server, just a different component)
- The JSP template engine picks up the JSP, and uses the model to perform server-side processing to generate an HTML
- The final HTML is sent to the client. It contains no JSP tags or even remnants of a model
- The client only interacts with this HTML. 
- If the HTML contains some calls to do things dynamically (like call other sources, including itself, such as via REST or websockets), these are now called, but this is out of the server's hand now. Now the client is in control.

The whole point of using websockets is that it allows the client to be dynamic and we don't have to go to the server again and again to generate html after every change. The client gets the HTML from the server once, and then just adapts based on the data that it retrieves from the server separately.

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
randomnicode: It is.

For now just add to at least the default locale.

Eventually we'll start using a free translation bot to update the rest of the files (weblate or something)

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
randomnicode: I don't think this is needed

airsonic-main/src/main/java/org/airsonic/player/controller/PodcastChannelController.java
```diff
@@ -53,13 +53,14 @@ protected ModelAndView handleRequestInternal(HttpServletRequest request, HttpSer
...
         ModelAndView result = new ModelAndView();
-        result.addObject("model", map);
 
         int channelId = ServletRequestUtils.getRequiredIntParameter(request, "id");
 
         map.put("user", securityService.getCurrentUser(request));
         map.put("channel", podcastService.getChannel(channelId));
         map.put("episodes", podcastService.getEpisodes(channelId));
+
+        result.addObject("model", map);
```
randomnicode, line 63: change not needed?

randomnicode:
So, before even this change, a few conceptual things...

Airsonic mostly operates on an MVC paradigm. The server generates the static HTML and that's it (occasionally DWR is used to communicate with the server)
Airsonic-Advanced modernizes this by utilizing websockets and massively transforming the pages. However, the key point to note here is that not all pages have been transformed.

podcasts are one of them. They still use the outdated MVC mechanism of population.
This is why you're having to do these gymnastics.

My suggestion here is maybe you could try converting it to websockets (and datatables) first?
If you want to retain it then there seem to be a whole lot going on here, and i'll try highlighting it

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
```diff
@@ -1,45 +1,148 @@
...
+~  GNU General Public License for more details.
+~
+~  You should have received a copy of the GNU General Public License
+~  along with Airsonic.  If not, see <http://www.gnu.org/licenses/>.
+~
+~  Copyright 2015 (C) Sindre Mehus
+--%>
+
+<html>
+<head>
```
randomnicode, line 23: unrelated change to pr

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
randomnicode: this bit isn't needed

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
```diff
@@ -70,28 +179,46 @@
...
+    &raquo; ${fn:escapeXml(model.channel.title)}</h1>
 <h2>
-    <span class="header"><a href="javascript:top.playQueue.onPlayPodcastChannel(${model.channel.id})"><fmt:message key="common.play"/></a></span>
+    <span class="header"><a href="javascript:top.playQueue.onPlayPodcastChannel(${model.channel.id})"><fmt:message key="main.playall"/></a></span>
 
     <c:if test="${model.user.podcastRole}">
         | <span class="header"><a href="javascript:deleteChannel()"><fmt:message key="common.delete"/></a></span>
-        | <span class="header"><a href="javascript:refreshChannels()"><fmt:message key="podcastreceiver.check"/></a></span>
+        | <span class="header"><a href="javascript:refreshChannels()"><fmt:message
+            key="podcastreceiver.check"/></a></span>
```
randomnicode, line 216: unnecessary change?

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
```diff
@@ -70,28 +179,46 @@
...
     </c:if>
 </h2>
 
-<div class="detail" style="padding-top:0.2em;white-space:normal;width:80%">${fn:escapeXml(model.channel.description)}</div>
+<div class="detail"
+     style="padding-top:0.2em;white-space:normal;width:80%">${fn:escapeXml(model.channel.description)}</div>
```
randomnicode, line 221: unnecessary change?

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
```diff
@@ -126,7 +252,7 @@
             </c:choose>
 
             <td class="truncate">
-                    <span title="${episode.title}" class="songTitle">${episode.title}</span>
+                <span title="${episode.title}" class="songTitle">${episode.title}</span>
```
randomnicode, line 255: unnecessary change?

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
```diff
@@ -138,16 +264,16 @@
...
+            <span class="detail">
+                <c:choose>
+                    <c:when test="${episode.status eq 'DOWNLOADING'}">
+                        <fmt:formatNumber type="percent" value="${episode.completionRate}"/>
+                    </c:when>
+                    <c:otherwise>
+                        <fmt:message key="podcastreceiver.status.${fn:toLowerCase(episode.status)}"/>
+                    </c:otherwise>
+                </c:choose>
+            </span>
```
randomnicode, line 276: unnecessary change?

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
```diff
@@ -166,21 +292,51 @@
...
 
-<table style="padding-top:1em"><tr>
-    <c:if test="${model.user.podcastRole}">
-        <td style="padding-right:2em"><div class="forward"><a href="javascript:downloadSelected()"><fmt:message key="podcastreceiver.downloadselected"/></a></div></td>
-        <td style="padding-right:2em"><div class="forward"><a href="javascript:deleteSelected()"><fmt:message key="podcastreceiver.deleteselected"/></a></div></td>
-    </c:if>
-    <td style="padding-right:2em"><div class="forward"><a href="javascript:refreshPage()"><fmt:message key="podcastreceiver.refresh"/></a></div></td>
-    <c:if test="${model.user.adminRole}">
-        <td style="padding-right:2em"><div class="forward"><a href="podcastSettings.view?"><fmt:message key="podcastreceiver.settings"/></a></div></td>
+<select id="moreActions" class="pagetype-dependent type-album type-video" onchange="actionSelected(this.options[selectedIndex].id);" style="margin-bottom:1.0em">
```
randomnicode, line 295: the classes look like they were copy pasted from mediaMain. they're not necessary here

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
```diff
@@ -166,21 +292,51 @@
...
+    <option id="top" selected="selected"><fmt:message key="playlist.more"/></option>
+    <option id="selectAll">&nbsp;&nbsp;<fmt:message key="playlist.more.selectall"/></option>
+    <option id="selectNone">&nbsp;&nbsp;<fmt:message key="playlist.more.selectnone"/></option>
+    <c:if test="${model.user.downloadRole}">
+        <option id="download">&nbsp;&nbsp;<fmt:message key="common.download"/></option>
```
randomnicode, line 300: This "Download" and the "Download" outside the select box is now confusing.

They both mean different types of downloads

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
randomnicode: needs to be put in a resource bundle...

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
```diff
@@ -70,28 +179,46 @@
...
 
 <div style="float:left;margin-right:1.5em;margin-bottom:1.5em">
-<c:import url="coverArt.jsp">
-    <c:param name="podcastChannelId" value="${model.channel.id}"/>
-    <c:param name="coverArtSize" value="200"/>
-</c:import>
+    <c:import url="coverArt.jsp">
+        <c:param name="podcastChannelId" value="${model.channel.id}"/>
+        <c:param name="coverArtSize" value="200"/>
+    </c:import>
```
randomnicode, line 205: ide doing funny stuff?

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
```diff
@@ -70,28 +179,46 @@
...
 </div>
 
-<h1 id="name"><a href="podcastChannels.view"><fmt:message key="podcastreceiver.title"/></a> &raquo; ${fn:escapeXml(model.channel.title)}</h1>
+<h1 id="name"><a href="podcastChannels.view"><fmt:message key="podcastreceiver.title"/></a>
+    &raquo; ${fn:escapeXml(model.channel.title)}</h1>
```
randomnicode, line 209: ide again

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
randomnicode: this function isn't necessary, and in fact, kind of is redundant with the function above, the additional filtering is only done for downloads, and you can just do that in the download block

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
```diff
@@ -59,7 +162,13 @@
...
         }
 
-        function getSelectedEpisodes() {
+        function selectAll(checked) {
+            for (var i = 0; i < ${fn:length(model.episodes)}; i++) {
+                $("#episode" + i).prop("checked", checked);
+            }
+        }
+
+        function getSelectedEpisodesId() {
```
randomnicode, line 171: i think maybe just keep the name to reduce noise

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
```diff
@@ -1,45 +1,148 @@
...
+            if (id == "top") {
+                return;
+            } else if (id == "selectAll") {
+                selectAll(true);
+            } else if (id == "selectNone") {
+                selectAll(false);
+            } else if (id == "download") {
+                if (getSelectedEpisodesMediaId().length > 0) {
+                    downloadMedia = true;
+                    getMediaDirectory(${model.channel.mediaFileId});
```
randomnicode, line 133: One, there is a security exposure here. Sure, in the UI its hidden behind a downloadrole, but in javascript, you're still giving folks the ability to at least ask the question.

Second, this has a race condition. The getSelectedEpisodesMediaId() might have changed between when the message was sent and when a reply was received (in getMediaDirectoryCallback())

Third, this seems like a very convoluted way to go about it. I think you're trying to retrieve the corresponding indices of the media file id in podcast channel directory? Wouldn't it be more prudent and straightforward to change the download controller to just accept multiple ids?

randomnicode:
Maybe try and reduce the noise due to ide formatting?

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
Faeb35: Thanks for the explanation. Helps a lot! I think I mainly already understand the idea behind Spring, creating the HTML on the server side and how websocket works in general.

My question was more pointing in another direction. I was asking myself how I can send a JavaScript object (object with some attributes) from the server side (Spring, JSP) to the client side (JavaScript).
Currently the most needed information about the episode are stored in the checkbox. The index in the html-id (not needed actually) and the episode id in the html-value of the checkbox. I had to get the status (if COMPLETED) and the mediaFileId (for adding to playlist and downloading as well). Currently the getSelectedEpisodes uses jQuery to get the id of the episode aka html-value from the checkbox. I think that's what you suggested for the status as well.

To get the mediaFileId and the status I saw the following options:
a) Adding additional pseudo HTML attributes to the checkbox or somewhere in the table to have the mediaFileId and the status-value (not text) in HTML (actually I did this first, but found it very odd). But I don't wanted to bloat HTML with information actually not needed there. And I also don't wanted to "parse" the HTML text by jQuery to get the status, as this one is translated.
b) Create a new PodcastWSController and start to port the whole podcast page to use DataTables and websockets, like you did for the album view, playqueue and the playlists. By using websockets I could easily pass a JavaScript object to the client side containing all the information needed. This was to much hustle for me, for one of my firs pull requests. I thought I could do this in a second iteration.
c) What I currently did: Generate JavaScript code on the server side to represent and create the JavaScript object needed. I found this way better then creating pseudo HTML attributes, bloating the HTML or go for the whole websockets implementation. Code on the client side looks like this now when rendered:

...          
var episode = {};
episode.id = '3';
episode.mediaFileId = '';
episode.status = 'SKIPPED';
podcastEpisodes.push(episode);

var episode = {};
episode.id = '2';
episode.mediaFileId = '402';
episode.status = 'COMPLETED';
podcastEpisodes.push(episode);

var episode = {};
episode.id = '7';
episode.mediaFileId = '';
episode.status = 'SKIPPED';
podcastEpisodes.push(episode);
...

Probably I could make this a bit slimmer by just generating a JSON string on the server side instead of a whole code. And yes, the id is not needed actually, but mediaFileId and status is.

Would be interesting to have your thoughts about this. I'm not quite sure how to handle that the best way.

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
Faeb35: I'll add the default local, that's fine.

You're right, this code is kind of redundant with the code above, but there are some differences which made it complicated to avoid the redundance for my at first glance. In getSelectedEpisodes I need the value and in getSelectedEpisodesMediaId I need just the index.

But when I think twice, I think I could make one funktion with this code (jQuery to get the checked episodes), returning a JavaScript array of objects, containing the index and the the values of the checkboxes. In getSelectedEpisodes I could use the value and in getSelectedEpisodesMediaId I could just use the index.

Also the getSelectedEpisodesMediaId was used 
Actually I

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
Faeb35: This existing "Download" button was very confusing for me since started using Airsonic. I thought that the "Download" would start a download to the client, but it actually starts Airsonic to download the podcast to the server. Related issue: https://github.com/airsonic/airsonic/issues/1479

What do you think about renaming the "Download" outside the select box to "Download to server"?

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
Faeb35: As I had the name getSelectedEpisodesMediaId to get the media Id, I tought it would make sense to have the Id there as well (as it actually returns ids).
I'm going to revert that.

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
Faeb35: Your absolutely right, "convoluted" sounds right :-) I had to use a lot of the websocket stuff, just to get the media id. I did all this for not needing to alter the download controller (noise reduction :-D), but actually it would have been better to just implement a change in the download controller ;-)

What do you think about if I would add a "p" parameter (instead of i) in the download controller which could point to the podcast episode id(s)?
In the download controller I would then need to get the mediaId's accordingly.

airsonic-main/src/main/java/org/airsonic/player/controller/PodcastChannelController.java
Faeb35: Absolutely unnecessary ;-) Going to revert that...

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
Faeb35: I'm going to remove the classes. And yes, it's just copy & paste (you got me).

Faeb35:
@randomnicode thank you very much for the review and patiently explaining me all the stuff. Hightly appreciated!
The most things I addressed already in the comments in the code above.

Sorry for all the mess with unintended and unnecessary code changes (tabs, spaces). I think I hit that "Reformat code"-button once in the IDE, which unfortunately rearranged some unnecessary things. I'm going to revert all that right now... Still beginnert, sorry for that!

I think the next main change will be to adjust the download controller to accept id's of podcast episodes beside mediaIds (if that's fine). By doing so I would get rid of a lot of the pain here ;-)
The only thing I'm not quite sure: How to get the download status of the episodes (and showing the error toast). Still using my current way of generating a JavaScript object on the server side or adding a pseude-attribute next to/on the staus column? Parsing the html text there doesn't sound feasible.

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
randomnicode: Right, I'm saying you don't need this.

You don't even need additional pseudo HTML attributes, because they already exist, and we can just use them.

Scenario: you have 2 objects selected and you want to find out (a) their mediaFileIds and (b) their statuses

Pseudocode:
- getSelectedEpisodes() is the key method here
- retrieve each checkbox ("#episode"+i) and see if it's selected (we already do this)
- if it is, that's the row we need
  - episode id is just the val() of the checkbox (we already know this and use it)
  - status is a td element in the same row
    - retrieve the row: parentRow = $(checkbox).parent('tr')
    - retrieve the text of the status td box inside the parent
      - use a class as a marker or the _nth_ child selector of the parent tr
         - statusbox = parentRow.children().eq(5) or statusbox = parentRow.find('.podcast-status'), are just examples of how you could engineer it
      - compare it with the localized translation to figure out the status
          - completedStatus = '<fmt:message key="podcastreceiver.status.completed}"/>'; statusbox.text().equals(completedStatus); --> COMPLETED
  - mediafileId is given to the playButtons.jsp. One of those elements has it embedded. Again, either use _nth_ child of parent or use a class
    - starRow = parentRow.children().eq(1) or add a class and use it to locate it
    - mediaFileId is then just the suffix of the id attribute (starImage123): mid = starRow.attr('id').substring(9) -> you can make it numeric if necessary

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
randomnicode: maybe:
- getSelectedEpisodes() returns an array of objects (or arrays) containing: {mediaId, id, status}
- everyone else filters this array prior to use to get the datapoints they need

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
randomnicode: sure, sounds good to me

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
randomnicode: I was thinking, you don't even need to add an additional parameter. Just make the id parameter accept lists. Then, since you already know mediafileids for all the selected episodes, pass those in directly!

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
Faeb35: Ok, got it! I think I will add a class to have a better and more reliable jQuery selector to locate the stuff.

**Status**: Compare with the translated version should do the trick!
**mediaFileId**: right, it's in the playButtons, but I need to be parsed out from the HTML 
Example: "Play last" button
`Play last`
==> I would take it from the id attribute as suggested

It's feasable, but I don't like "parsing" the HTML and it's attribute, as we would need to alter the selector when there's a change in the playButtons.jsp. Just sounds not very reliable to me. But anyway, I'll do it for now. When we have websockets we will have better options ;-)

airsonic-main/src/main/webapp/WEB-INF/jsp/podcastChannel.jsp
Faeb35: Had to stare at the code for I while to get it :-)
Got it... finally ;-)

randomnicode:
```
function getSelectedEpisodes() {
            var result = [];
            for (var i = 0; i < ${fn:length(model.episodes)}; i++) {
                var checkbox = $("#episode" + i);
                if (checkbox.is(":checked")) {
                    var id = checkbox.val();
                    var parent = checkbox.parent('tr')l
                    var mediaFileId = parent.children().eq(1).attr('id').substring(9);
                    var downloaded = parent.children().eq(somecellindex).text().equals('<fmt:message key="podcastreceiver.status.downloaded"/>');
                    result.push({stuff-above});
                }
            }
            return result;
        }
```

airsonic-main/src/main/java/org/airsonic/player/controller/DownloadController.java
```diff
@@ -103,7 +103,7 @@
 
     @GetMapping
     public ResponseEntity<Resource> handleRequest(Principal p,
-            @RequestParam Optional<Integer> id,
+            @RequestParam(name = "id") Optional<List<Integer>> ids,
```
Faeb35, line 106: I changed this to a List but still using the Optional. But I'm not quiet sure if I could change this to (required = false) and check for null instead like the other parameters. what's the difference?

randomnicode: No difference. Just in the coding style that you want to adopt or pursue later on in the method (imperative or functional, or even declarative). Monads like the Optional help you pursue the latter, required = false are more aligned to the former.

randomnicode:
Sorry, I haven't reviewed this in a bit. I hope to soon.

PR #302, opened 2020-05-14
Fix getting started page truncated

Faeb35:
Before:
grafik

After:
grafik


randomnicode:
So bad.

Thank you!

randomnicode:
This PR seems incomplete and missed these 4 locales:
de
es
nl
zh_TW

PR #303, opened 2020-05-14
Save "add to playlist"-dialog size

Faeb35:
This pull request adds the ability to save the size of the dialog, showing when adding tracks to a playlist. I found it very annoying to change the size of this dialog every time.

The size is saved in cookies. This makes it very lightweight and also adds the possibility to have different sizes on different computers/browsers for the same user/player.

The size of the dialog, shown when adding from the album view, is stored in a different cookie then the size of the dialog shown, when adding from the player queue. Even though the initial size is 300 width and 250 height for both (height was 350 for mediaMain and 220 for playQueue).

### DIFF
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/mediaMain.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/mediaMain.jsp
@@ -245,12 +245,15 @@
             minBorder: 30
         });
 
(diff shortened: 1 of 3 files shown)


Faeb35:
This pull request is on hold until reviewed and merged here: https://github.com/airsonic/airsonic/pull/1714

PR #307, opened 2020-05-15
Fix showing album view for podcasts

Faeb35:
Bug was reported here by myself: https://github.com/airsonic/airsonic/issues/1554

When clicking on several links of podcasts, you're getting to the album view of podcast, which actually shouldn't exist. You should get the podcast channel view.
This pull requests fixes all these links by forwarding from main.view to the corresponding podcastChannel.view. I found this approach way easier then changing all the links around in Airsonic (now playing/recently played, playlist and player).

Unfortunately there was no real link from the MEDIA_FILE table to the PODCAST_EPISODE table to get the podcast channel. The only link was the PATH column. So I had to use this column to get the corresponding podcast channel of the media file id passed by the url in the MainController.
I think it would be better to have a propper key between MEDIA_FILE and PODCAST_EPISODE, but adding doing that is way beyond my skills.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/MainController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/MainController.java
@@ -28,6 +28,7 @@ import org.springframework.web.bind.ServletRequestUtils;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.servlet.ModelAndView;
+import org.springframework.web.servlet.view.RedirectView;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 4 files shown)


airsonic-main/src/main/java/org/airsonic/player/controller/MainController.java
```diff
@@ -47,13 +48,22 @@
...
 
     @GetMapping
     protected ModelAndView handleRequestInternal(HttpServletRequest request, HttpServletResponse response) throws Exception {
         Map<String, Object> map = new HashMap<>();
 
         String username = securityService.getCurrentUsername(request);
         UserSettings userSettings = settingsService.getUserSettings(username);
+        int[] requestParameterIds = ServletRequestUtils.getIntParameters(request, "id");
+
+        if (requestParameterIds != null && requestParameterIds.length == 1 && isPodcast(requestParameterIds[0])) {
```
Faeb35, line 64: I'm not quiet sure, why we could have multiple ids set in the url. I tested it out, it's working, but I've never seen this in "wildlife". To make this fix not overcomplicated (ids could be mixed with albums and podcast channels), I'm only handling it when there's only one id.

PR #309, opened 2020-05-16
Update InternetRadioService to support direct binary urls

Faeb35:
Fixes this issue https://github.com/airsonic/airsonic/issues/1209
In the current release the specified radio url in the settings should point to a m3u-file containing urls to radio streams.
With this update you can now also directly specify the url to a binary radio stream.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/InternetRadioService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/InternetRadioService.java
@@ -155,10 +155,29 @@ public class InternetRadioService {
      * @return a list of internet radio sources
      */
     private List<InternetRadioSource> retrieveInternetRadioSources(InternetRadio radio, int maxCount, long maxByteSize, int maxRedirects) throws Exception {
-        // Retrieve the remote playlist
-        String playlistUrl = radio.getStreamUrl();
-        LOG.debug("Parsing internet radio playlist at {}...", playlistUrl);
-        SpecificPlaylist inputPlaylist = retrievePlaylist(new URL(playlistUrl), maxByteSize, maxRedirects);
+        String streamUrl = radio.getStreamUrl();
+        LOG.debug("Parsing internet radio (playlist) at {}...", streamUrl);
+
+        SpecificPlaylist inputPlaylist = null;
+        HttpURLConnection urlConnection = connectToURLWithRedirects(new URL(streamUrl), maxRedirects);
+        try (InputStream in = urlConnection.getInputStream()) {
+            String contentType = urlConnection.getContentType();
+            if ("audio/mpeg".equals(contentType)) {
+                //for direct binary streams, just return a collection with a single internet radio source
+                return Collections.singletonList(new InternetRadioSource(streamUrl));
+            }
+            String contentEncoding = urlConnection.getContentEncoding();
+            if (maxByteSize > 0) {
+                inputPlaylist = SpecificPlaylistFactory.getInstance().readFrom(new BoundedInputStream(in, maxByteSize), contentEncoding);
+            } else {
+                inputPlaylist = SpecificPlaylistFactory.getInstance().readFrom(in, contentEncoding);
+            }
+        } finally {
+            urlConnection.disconnect();
+        }
+        if (inputPlaylist == null) {
+            throw new PlaylistFormatUnsupported("Unsupported playlist format " + streamUrl);
+        }
 
         // Retrieve stream URLs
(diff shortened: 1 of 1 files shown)


airsonic-main/src/main/java/org/airsonic/player/service/InternetRadioService.java
```diff
@@ -130,19 +130,32 @@ public void clearInternetRadioSourceCache(Integer internetRadioId) {
...
+            return entries;
+        } else {
+            //for playlists containing binary streams
+            return retrieveInternetRadioSources(
+                    radio,
+                    PLAYLIST_REMOTE_MAX_LENGTH,
+                    PLAYLIST_REMOTE_MAX_BYTE_SIZE,
+                    PLAYLIST_REMOTE_MAX_REDIRECTS
+            );
+        }
```
randomnicode, line 157: I think this has 2 issues:
1. It's connecting twice. You connect to the URL in L142, but then will connect again inside `retrieveInternetRadioSources(...)`
2. There is a connection leak. The urlConnection in L142 is never disconnected or released.

I suggest
- collapse retrievePlaylist() and combine it with retrieveInternetRadioSources(radio, maxCount, ...) 
- do this business inside the try-finally of the (former) retrievePlaylist()

something like:
```
private List<InternetRadioSource> retrieveInternetRadioSources(InternetRadio radio, int maxCount, long maxByteSize, int maxRedirects) throws Exception {
        // Retrieve the remote playlist
        String playlistUrl = radio.getStreamUrl();
        LOG.debug("Parsing internet radio playlist at {}...", playlistUrl);
        SpecificPlaylist inputPlaylist = null;
        HttpURLConnection urlConnection = connectToURLWithRedirects(url, maxRedirects);
        try (InputStream in = urlConnection.getInputStream()) {
            String contentType = urlConnection.getContentType();
            if ("audio/mpeg".equals(contentType)) {
                //for direct binary streams
                return Collections.singletonList(new InternetRadioSource(streamUrl));
            }
            String contentEncoding = urlConnection.getContentEncoding();
            if (maxByteSize > 0) {
                inputPlaylist = SpecificPlaylistFactory.getInstance().readFrom(new BoundedInputStream(in, maxByteSize), contentEncoding);
            } else {
                inputPlaylist = SpecificPlaylistFactory.getInstance().readFrom(in, contentEncoding);
            }
        } finally {
            urlConnection.disconnect();
        }
        if (inputPlaylist == null) {
            throw new PlaylistFormatUnsupported("Unsupported playlist format " + url.toString());
        }
        // Retrieve stream URLs
        List<InternetRadioSource> entries = new ArrayList<>();
        ...
```

```

**Comment by Faeb35** [comment]
@randomnicode thank you very much for the review, your time and pointing out the weak parts, highly appreciated! 
I just updated the pull request accordingly.
- retrieveInternetRadioSources(InternetRadio radio) changes reverted
- merged the retrievePlaylist into retrieveInternetRadioSources(radio, maxCount, ...) 
- added 
```if ("audio/mpeg".equals(contentType)) {
                //for direct binary streams, just return a collection with a single internet radio source
                return Collections.singletonList(new InternetRadioSource(streamUrl));
            }
```

Nice to see, that you converted 
```
            List<InternetRadioSource> entries = new ArrayList<>();
            entries.add(new InternetRadioSource(streamUrl));
            return entries;
```
to
`Collections.singletonList(new InternetRadioSource(streamUrl));`
Way smaller and better readable!
Learned a new trick ;-) I guess there are some more to come (as I'm still beginner with Spring, Git and never used Java for like 4 years).

randomnicode:
>Way smaller and better readable!
>Learned a new trick ;-) I guess there are some more to come (as I'm still beginner with Spring, Git and never used Java for like 4 years).

Tradeoffs for everything. The tradeoff here is _mutability_. Whoever you return this to, won't be able to mutate the returned list anymore (which you can with the other implementation).

LGTM is whining about the BoundedInputStream leaking. Since we always have a maxByte, do you want to fix that? I suggest take out the `if..else` predicated on `maxBytes > 0` (because it'll always be > 0 in this context) and just move the BoundedInputStream constructor to the try.

Something like:

```
try (InputStream in = .... ; BoundedInputStream bin = new BoundedInputStream(in, maxByteSize);) {
...
  String contentEncoding = urlConnection.getContentEncoding();
  playlist = SpecificPlaylistFactory.getInstance().readFrom(bin, contentEncoding);
} finally {
...
}
```

Faeb35:
Thanks for the LGTM update and merging!

PR #315, opened 2020-05-20
Introduce exploded and layered jar in container

randomnicode:
This PR fundamentally changes the way the docker images are packaged for this repo.
They used to be a simple war file executed inside the (JRE) container.

This PR changes this fact in two ways:
- The war is exploded and its contents are uncompressed
- The contents are _layered_ into the image, using best estimates of frequency of changes

Pros:
- The expanded classes mean _at least_ startup is much faster. Cursory comparisons via the tests prior and in this PR show startup times reduce from ~21s to ~14s. Other portions may be faster also. Spring's jar classloader is no longer needed or used.
- The layers means that the whole image doesn't have to be pulled when building or deploying over and over again, especially at update. Dependencies are relatively static as opposed to application code. If just the application code changes, dependencies exist in a different layer, and don't have to be repulled or rebuilt.

Cons:
- Exploding jars is expensive and the image takes about ~1-2 min longer to build (one time cost incurred by PR test or at merge)
- Static initial space usage might be possibly higher due to more layers (even though you pull down less bytes when updating)? Unsure


Rudd-O:
I confirm that the exploded JAR works properly by having reproduced the exact same explosion process in my RPM packaging, and verifying it (1) works (2) starts twice as fast.

Rudd-O:
Thumbs up on merge in my view.

PR #317, opened 2020-05-22
Add option to choose cover art source and add option for full scan

Faeb35:
This pull requests add two options to settings menu. 

**Option to select the source where the cover art should be taken from**
grafik
(image updated after last commit)

In the current release Airsonic always tries to search the cover as file in the folder. If there's no cover it will try to take the cover embedded in the tags from the first file in the folder.
For me this wasn't sufficient. Mostly I've my covers correctly stored in the tags, but I've multiple images in the folder. So Airsonic ended up to just choose the first image in the folder, which is not always the right one. (e.g. when I have "back.jpg" and "front.jpg" always "back.jpg" was taken).

With this options it's now possible to have the covers in the tags prefered over the covers files in the folder. If someone would like to have covers only from files or only from tags, I added these options as well.

But unfortunately after changing this setting, Airsonic didn't update my covers. I discovered several related issues about this and that there was a propperty added to force Airsonic to scan all the files again https://github.com/airsonic/airsonic/pull/607 This worked like a charm, but I don't think that all the users would like to add a property to the airsonic.properties file after changing this setting. So I added an option for that as well to the settings.

**Option to force a full rescan**
grafik

This option not only helps for rescanning the medias after changing the new introduced option with this pull request. It also should address and help with several other issues.

The following open issues are (probably) related to this pull request:
https://github.com/airsonic/airsonic/issues/649
https://github.com/airsonic/airsonic/issues/711
https://github.com/airsonic/airsonic/issues/732
https://github.com/airsonic/airsonic/issues/359
https://github.com/airsonic/airsonic/issues/797


Faeb35:
I'm not quite sure if we could rename the already existing property `IgnoreFileTimestamps` to something like `doFullScan`. "IgnoreFileTimestamps" is more technically describing what it's doing, but I think it also would make sense to call it more from the functional perspective, as "doFullScan"  is what it's doing in the end.

airsonic-main/src/main/java/org/airsonic/player/command/GeneralSettingsCommand.java
```diff
@@ -81,6 +82,14 @@ public void setCoverArtFileTypes(String coverArtFileTypes) {
         this.coverArtFileTypes = coverArtFileTypes;
     }
 
+    public String getCoverArtSource() {
```
randomnicode, line 85: I'd prefer this was an enum, just to restrict the choices?

airsonic-main/src/main/java/org/airsonic/player/controller/GeneralSettingsController.java
```diff
@@ -105,13 +106,18 @@ protected String doSubmitAction(@ModelAttribute("command") GeneralSettingsComman
 
         redirectAttributes.addFlashAttribute("settings_toast", true);
 
+        if (!settingsService.getCoverArtSource().equals(command.getCoverArtSource())) {
+            settingsService.setIgnoreFileTimestamps(true);
```
randomnicode, line 110: yeah, agreed, this is a terrible name

airsonic-main/src/main/java/org/airsonic/player/service/MediaFileService.java
```diff
@@ -586,6 +586,30 @@ public Path getCoverArt(MediaFile mediaFile) {
      * Finds a cover art image for the given directory, by looking for it on the disk.
      */
     private Path findCoverArt(Collection<Path> candidates) {
+        Path candidate = null;
+        var coverArtSource = settingsService.getCoverArtSource();
+        if ("tag+file".equals(coverArtSource)) {
```
randomnicode, line 591: i'd prefer these as enums

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -276,6 +276,12 @@ private void doScanLibrary(ForkJoinPool pool) {
 
             LOG.info("Completed media library scan.");
 
+            if (settingsService.isIgnoreFileTimestamps()) {
+                settingsService.setIgnoreFileTimestamps(false);
+                settingsService.save();
+                LOG.info("Ignore file timestamp property (for full scan) reset");
+            }
```
randomnicode, line 283: Hmmm, there is a distinction to be made here between a one-time setting and a setting that stays scan-after-scan.
You're overriding it here.

Say a person wants to do a full scan every time, and sets the property in his prop file. This will override it (set it to false).

randomnicode:
So here are a few comments:
Logistics:
- You don't need to add the i18n strings for all the ResourceBundles in your PR. It just artificially bloats the PR. I think the PR should focus on the actual issue. i18n is a separate issue and should be addressed in a separate PR. (So just add strings to the default ResourceBundle and that should be enough in this repo).

The PR itelf:
- I think we should make the sourcetype an enum
- ignoreFileTimestamps is a terrible property name and we should change it. It doesn't even have context as to what part of airsonic it is referring to (scans). I think "fullScan" or "detailedScan" sounds fine (I prefer the latter). Additionally, we should explain in the UI (wherever we surface it) what it means (like in a hover tip). Something like "Normally scanning checks disk file timestamps and if they match with the db, the data in the db for the file is not changed or updated; the scanner ignores the file and moves on to the next file. This property forces the scanner to recreate and update the data for each file regardless of the disk timestamp" 
- We need to make a distinction between one-time detailed scan (what you're trying to do here), and a recurring detailed scan (which already exists if you set the property in your property file). I think it should be done in a separate PR. I think you need another property for that (like ClearDetailedScanSettingOnScanComplete). On scan complete, you check the flag, and then clear the flag and the setting if the flag is set. The UI should be something like:
  - [checkbox] Enable detailed scan (hover tip detail)
        - [checkbox] Next time only
- Then you can bolt on your scan requirement. The logic: if detailed scan is NOT set, then set it and the clear-after-scan flag, otherwise do nothing (the next scan will already be a detailed scan because user has already set it)


So, in short, if you're up for it, what I propose is:
- A PR for renaming the IgnoreTimestamp
- Stack a PR on top of it (just build another branch on top of the previous commit) for adding another setting flag for clearing the previous setting on scan completion
- Stack another PR for exposing the two in the UI
- Stack this PR on top using the capabilities in the previous PRs

Or you can just do the requested changes without the scanning portion for this PR and do the scanning bit later in follow up PRs?

Faeb35:
Thanks for your review and inputs! Highly appreciated and nice to get a feedback from someone more experienced with all that stuff here ;-)

- I defenitly will introduce a enum for the sourcetype and use switch-case.
- I'll rename everything like "doScan" and the "IgnoreFileTimestamps"-property to "fullScan" (property renaming will be done in an own PR). I would prefere the word "full" -> scanning the full libary, "detailed" sounds to me like it would scan additional details from the files or collection some more details from the folders.
- About the hover tip: I'm going to add the suggested information, about what a full scan actually means to the already existing hover tip there.
- But I don't agree that we should have distinct between on-time full scan and recurring full scan. I see your point, that it could be a bit unexpected, when the property get's reset, but for that I added the log entry to make it more obvious. I think Airsonic should usually work just fine only scanning the updated files as it already does on every scan. Only under some special circumstances its needed to have a full scan. As I've seen in the related issues and the original pull request about the  "IgnoreFileTimestamps"-property, this setting was only been to come over these special circumstances, where the cover or the library was not updating or after a restore of media folders. If we would add a second option to have the possibility to enabled full scan all the time, I'm afraid that people could turn this on accidentaly, think that a non-full scan wouldn't be sufficiant. The property was only made to be set once and to be used very seldomly. That's why I chose the behaviour of reseting it every time. Personally I also don't like having to many settings as it can be rather confusing for user not understanding all the details. I'm more for the "keep it simple" approach and having one option which is reseting. Do you really think that somebody would need to have the full library scanned every time? I don't see a real need for that. It's only needed with one of the named special circumstances and even then only once.

About stacking PR or doing the scanning in another PR:
I bundled the full scan setting with the cover art source setting, as changing cover art source really needs a full rescan and isn't working without this. I was thinking that these two changes need to come at the same time. Only option I see to split it, we would need to explain how to handle the "IgnoreFileTimestamps"-property in the hover tip with the cover art source PR. In the second PR (for the full scan PR) this would need to be removed again.
What about keep this PR as it is (full scan and cover source bundled) and doing the renaming of the property in a later PR?

About i18n: I'll only add the stuff to the ResourceBundle_en.properties in future. Glade to save that work ;-)
Should I revert the changes to the other i18n files in this PR?

Faeb35:
I just pushed a new commit, which includes an enum for the cover art source instead of a string. I added the enum to the domains folder, hope that's the right place...
In the SettingsService I tried to convert the cover art source property to an enum. If the current property isn't a proper enum value I added a try catch and returning the default value. Not quit sure if it's best practice to handle that like this.
Additionally I renamed all of the "doFullScan" things to "fullScan".

randomnicode:
> But I don't agree that we should have distinct between on-time full scan and recurring full scan. I see your point, that it could be a bit unexpected, when the property get's reset, but for that I added the log entry to make it more obvious.

There is NO precedence  and nor should we make one of a system overwriting a user-specified property.
It should not happen because it leads to unexpected situations where a user specifically sets a property, and a system deliberately overrides the user wishes. If a user wishes to NOT do the full scan, s/he should be the one to specify the revert themselves too.

>  I think Airsonic should usually work just fine only scanning the updated files as it already does on every scan.

Yep, sure. It should.

>  Only under some special circumstances its needed to have a full scan.

Yep, agreed. Special circumstances SPECIFIED by the user, and therefore the _end_ of special circumstances should be specified by the user too.

> As I've seen in the related issues and the original pull request about the "IgnoreFileTimestamps"-property, this setting was only made to come over these special circumstances, where the cover or the library was not updating or after a restore of media folders. If we would add a second option to have the possibility to enabled full scan all the time

I think you misunderstood. We're not ADDING the possibility to enable full scan all the time. The current property already DOES THAT; it enables it "all the time" unless you manually revert it. We're RETAINING existing behavior with the existing property.
The second property you'd be adding ADDs the possibility to do the one-time full-scan, which _is_ a NEW feature.

>  I'm afraid that people could turn this on accidentaly, think that a non-full scan wouldn't be sufficiant.

Good thing we have a hover tip disseminating the information! Plus we have defaults, forums etc etc.

>  I'm more for the "keep it simple" approach and having one option which is reseting. Do you really think that somebody would need to have the full library scanned every time? I don't see a real need for that. It's only needed with one of the named special circumstances and even then only once.

My point of view is consistency. You're suddenly changing the behavior of the system here with regards to a particular property. Previously someone who wanted to have it enabled all the time used to be able to do it. Now his system keeps resetting it suddenly.

Furthermore, taking a step further, why does this not happen with other properties? Reset them after every use? Clearly there is something special about this property.

In general, a property set shouldn't be reset UNLESS you are telling the system to reset it. That's what the second property is doing: telling the system to reset.

> I bundled the full scan setting with the cover art source setting, as changing cover art source really needs a full rescan and isn't working without this

I think you're changing the dependencies here. Cover art source needs a full rescan, sure. But does a full rescan need a cover art source? No. Therefore it can be independent. Does a one-time full rescan need a cover art source? No. Independent. Does one-time full rescan need full rescan yes? okay stack. And so on. Which is why I presented the stacked PR order in a particular order.

> Should I revert the changes to the other i18n files in this PR?

Sure, up to you.

randomnicode:
superseded by #345

PR #348, opened 2020-07-14
Fix player id autoincrement issue

randomnicode:
New players cannot be created because the code now relies on autoincrement feature for ids in player table but that schema modification wasn't done yet.


randomnicode:
already autogenerated?

Issue #359, opened 2020-07-31
Arm/Arm64 Docker Build


TheFlatworm:
Would it be possible to add an Arm/Arm64 docker build?

randomnicode:
Hey @TheFlatworm , want to submit some PRs for it? We're happy to add them

TheFlatworm:
I'm a noob but looking into how to do this!

jokeofweek:
I believe this should be fairly straightforward... would likely need a second Dockerfile (Dockerfile.armhf) and it should pull from a different root image.

Instead of

```
FROM adoptopenjdk:14-jre-hotspot
```

It should be

```
FROM arm32v7/adoptopenjdk:14-jre-hotspot
```

yllekz:
Hi folks, does this exist? I'd like to deploy this to my Pi but I can't find an arm tag.

randomnicode:
Since tini has now been taken out from the Dockerfile, you could build on your machine with buildx for crossplatform support.

https://github.com/airsonic-advanced/airsonic-advanced/pull/438 might provide a hint on how to do it

If you have support to build linux/arm64 (qemu and buildx):
-build airsonic-advanced (or put the war in install/docker/target/dependency as airsonic-main.war)
- arm64 run: docker buildx --platform linux/arm64 --load -f install/docker/Dockerfile -t airsonicadvanced/airsonic-advanced:arm64-latest install/docker
- armv7 run: docker buildx --platform linux/arm/v7 --load -f install/docker/Dockerfile -t airsonicadvanced/airsonic-advanced:armv7-latest install/docker

nesasio93:
Building the image is something not everyone will be capable to do, especially when buildx is a experimental feature of docker. Are there plans to officially support arm/arm64? I would love to try this out on my RPI4! 😄

yllekz:
Same honestly. I'm not a developer and I can't just "build it on my own machine" honestly.

randomnicode:
> Building the image is something not everyone will be capable to do, especially when buildx is a experimental feature of docker.

> Same honestly. I'm not a developer and I can't just "build it on my own machine" honestly.

Fair enough.

okay, in the interest of expediting this, we can roll out an arm build, however it won't have gone through the same testing process. The war itself will be tested as part of the ci/cd, but the container image (for arm builds) may not be unless someone wants to contribute PRs

nesasio93:
[...]

PR #362, opened 2020-08-20
Add armhf Dockerfile.

jokeofweek:
This adds an armhf equivalent of the Dockerfile, addressing https://github.com/airsonic-advanced/airsonic-advanced/issues/359 (though you'd likely need something similar for 64 bit). This lets you run airsonic-advanced on devices like Raspberry Pi.

The only differences are the adoptopenjdk and the tini binary to use.

The docker manifest would need to be updated to publish both architectures (eg. https://dev.to/toolboc/publish-multi-arch-docker-images-to-a-single-repo-with-docker-manifests-329n)

### DIFF
--- a/install/docker/Dockerfile.armhf
+++ b/install/docker/Dockerfile.armhf
@@ -1,0 +1,44 @@
+FROM arm32v7/adoptopenjdk:14-jre-hotspot
+
+LABEL description="Airsonic-Advanced is a free, web-based media streamer, providing ubiquitous access to your music." \
+      url="https://github.com/airsonic-advanced/airsonic-advanced"
+
+ENV AIRSONIC_PORT=4040 AIRSONIC_DIR=/var CONTEXT_PATH=/ UPNP_PORT=4041 PUID=0 PGID=0
+
+WORKDIR $AIRSONIC_DIR
+
+ADD https://github.com/krallin/tini/releases/download/v0.19.0/tini-armhf /usr/local/bin/tini
+RUN chmod +x /usr/local/bin/tini
+
+RUN apt-get update && \
+    apt-get install -y software-properties-common && \
+    add-apt-repository -y ppa:jonathonf/ffmpeg-4 && \
+    apt-get update && \
+    apt-get install -y ffmpeg \
+                       x264 \
+                       x265 \
+                       lame \
+                       bash \
+                       ttf-dejavu \
+                       gosu
+
+COPY run.sh /usr/local/bin/run.sh
+RUN chmod +x /usr/local/bin/run.sh
+COPY entry.sh /usr/local/bin/entry.sh
+RUN chmod +x /usr/local/bin/entry.sh
+
+COPY target/dependency/airsonic-main.war airsonic.war
+
+EXPOSE $AIRSONIC_PORT
+
+# Default DLNA/UPnP ports
+EXPOSE $UPNP_PORT
+EXPOSE 1900/udp
+
+USER ${PUID}:${PGID}
+
+VOLUME $AIRSONIC_DIR/airsonic $AIRSONIC_DIR/music $AIRSONIC_DIR/playlists $AIRSONIC_DIR/podcasts
+
+HEALTHCHECK --interval=15s --timeout=3s CMD curl -f http://localhost:"$AIRSONIC_PORT""$CONTEXT_PATH"rest/ping || false
+
+ENTRYPOINT ["tini", "--", "entry.sh"]


jokeofweek:
Would likely want to have a build test for this file as well - not sure the right place to update CI.

randomnicode:
Haven't reviewed the PR yet, but this is great!

> Would likely want to have a build test for this file as well - not sure the right place to update CI.

That would be great. The CI actually sits here: https://github.com/airsonic-advanced/airsonic-advanced/tree/master/.github/workflows

But that's not sufficient by because the image is built here: https://github.com/airsonic-advanced/airsonic-advanced/blob/44f4214df92b200f694b566241359fd299a21620/install/docker/pom.xml#L29-L49

and tested here

https://github.com/airsonic-advanced/airsonic-advanced/blob/44f4214df92b200f694b566241359fd299a21620/integration-test/pom.xml#L121-L156

So we'll likely need to rework the test harness a bit (along with the CI to provide the correct host arch).

randomnicode:
PS have you tested this locally to see if it works?

jokeofweek:
Yep, confirmed that this works (using it on my raspberry pi as we speak 😄).

jokeofweek:
I played around with this a bit more - really the only actual issue in this Dockerfile is the `tini` binary.

As of Docker 1.13, `tini` is built-in so you actually don't need to use this anymore. See https://github.com/krallin/tini for details.

If you want to make just one Dockerfile that is platform-agnostic, the following Dockerfile would do the trick:

```
FROM adoptopenjdk:14-jre-hotspot

LABEL description="Airsonic-Advanced is a free, web-based media streamer, providing ubiquitous access to your music." \
      url="https://github.com/airsonic-advanced/airsonic-advanced"

ENV AIRSONIC_PORT=4040 AIRSONIC_DIR=/var CONTEXT_PATH=/ UPNP_PORT=4041 PUID=0 PGID=0

WORKDIR $AIRSONIC_DIR

RUN apt-get update && \
    apt-get install -y software-properties-common && \
    add-apt-repository -y ppa:jonathonf/ffmpeg-4 && \
    apt-get update && \
    apt-get install -y ffmpeg \
                       x264 \
                       x265 \
                       lame \
                       bash \
                       ttf-dejavu \
                       gosu

COPY run.sh /usr/local/bin/run.sh
RUN chmod +x /usr/local/bin/run.sh
COPY entry.sh /usr/local/bin/entry.sh
RUN chmod +x /usr/local/bin/entry.sh

COPY target/dependency/airsonic-main.war airsonic.war

EXPOSE $AIRSONIC_PORT

# Default DLNA/UPnP ports
EXPOSE $UPNP_PORT
EXPOSE 1900/udp

USER ${PUID}:${PGID}

VOLUME $AIRSONIC_DIR/airsonic $AIRSONIC_DIR/music $AIRSONIC_DIR/playlists $AIRSONIC_DIR/podcasts

HEALTHCHECK --interval=15s --timeout=3s CMD curl -f http://localhost:"$AIRSONIC_PORT""$CONTEXT_PATH"rest/ping || false

ENTRYPOINT ["entry.sh"]
```

For the same behavior, you have to either use `docker run --init` or if using docker-compose, put:

```
services:
  airsonic-advanced:
    init: true
```

This might be cleaner in the long run - easier to just maintain one good platform-agnostic Dockerfile.

This would also fix the issue for arm64 (https://github.com/airsonic-advanced/airsonic-advanced/pull/363)

Let me know if you prefer this and I can change my pull request.

randomnicode:
> I played around with this a bit more - really the only actual issue in this Dockerfile is the tini binary.

Hmm, what about the JRE?

> This might be cleaner in the long run - easier to just maintain one good platform-agnostic Dockerfile.

I also think this is a good idea, if this is indeed possible.

But we should do a stable release first, because it requires a non-backwards compatible change in the invocation and docker versions. Maybe this weekend we can do a stable release since the build has been pretty stable? Then we can merge in your PR into master?

Can you make the requisite changes in this PR? You might need to change the invocation for tests also. And maybe add a note to the README.md in the differences?

- [ ] Update PR to make unified Dockerfile
- [ ] Change test invocation to ensure container comes up fine during tests
- [ ] Update README.MD and add a note explaining the new invocation for bringing up containers
- [ ] Publish images of multiple arch/OS to Docker with manifests (separate PR)

jokeofweek:
The correct JRE will be picked. The main issue was that the tini executable being pulled in was compiled for x86. This resulted in a common docker error when running the binary on an ARM device: 

```
standard_init_linux.go:211: exec user process caused "exec format error"
```

I can get to this in the next few days ideally.

My main question is: was there a reason you were relying on tini explicitly? Since it's not actually the default setting and must be set with `--init` (https://docs.docker.com/engine/reference/run/#specify-an-init-process), want to check with you that you actually want to keep this around. Was there historically an issue in the past with things like zombie processes?

randomnicode:
> My main question is: was there a reason you were relying on tini explicitly?

Yeah, good point.

Seems like tini was pulled into Dockerfile in this commit: https://github.com/airsonic-advanced/airsonic-advanced/commit/e1ecc2444cb120c6c5cafe0a4c2926fa14368fe5#diff-87ba2d2b0c6cfbf4eb068c7b4a8e1579R29, but was historically being used even before that:
https://github.com/airsonic-advanced/airsonic-advanced/commit/e1ecc2444cb120c6c5cafe0a4c2926fa14368fe5#diff-8f6f903a1d44dd0e655bfd12a248b658L9

No idea whether there was an issue with zombies.

install/docker/Dockerfile.armhf
```diff
@@ -0,0 +1,44 @@
+FROM arm32v7/adoptopenjdk:14-jre-hotspot
```
randomnicode, line 1: this file isn't platform-agnostic?

jokeofweek:
I've dug into this a bit more.

Unfortunately I think removing tini is risky as:
- The maven plugin you're using doesn't support instantiating with init (see https://github.com/fabric8io/docker-maven-plugin/issues/1003) so the integration tests wouldn't be able to run using `--init`.
- Completely removing the tini integration could mean that it can cause issues when running multiple services in the same container (https://docs.docker.com/config/containers/multi-service_container/). While there may not be any problems with airsonic-advanced, there definitely could be and I don't want to cause issues for users.

So I think sticking to platform-specific Dockerfiles might be a safer bet for now.

Unfortunately I'm not very familiar with Maven / couldn't get the set up to work locally, so I'm not sure how much I can help refactor the test harness. Going to leave this PR as is in case anyone else wants to run this on a Pi/ARM device. Sorry about the trouble!

yllekz:
Hi folks, is there a tag for this build? Looking to upgrade from airsonic to airsonic-advanced on my Pi but I can't find a tag that works. I read through this thread but am not sure what to use.

randomnicode:
since the current releases are now multiplatform as of https://github.com/airsonic-advanced/airsonic-advanced/pull/463, may we close this PR?

PR #392, opened 2020-10-07
Put radio station names in title attribute

Efreak:
This should allow seeing the radio station name if it doesn't fully fit in the view.

This is *not* tested, as I do not have an environment to test in, however it's fairly simple and should have no problems as long as quotes are being escaped somewhere (they generally aren't be in URLs anyways)

Contributing
  1.  **License Acceptance** All contributions must be licensed as [GNU GPLv3](https://github.com/airsonic/airsonic/blob/develop/LICENSE.txt) to be accepted. Use [`git commit --signoff`](https://jk.gs/git-commit.html) to acknowledge this.

I'm in the browser, and I'm pretty sure this is trivial enough to not need it? I can (eventually) clone locally if necessary, but my current environment is broken due to a partially restored backup.


randomnicode:
looks good, thanks

PR #417, opened 2020-12-23
rest api: add support for basic auth

tamland:
Adds support for using HTTP basic authentication with REST API instead of passing credentials in query string.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
@@ -34,6 +34,7 @@ import org.springframework.security.crypto.password.Pbkdf2PasswordEncoder;
 import org.springframework.security.crypto.scrypt.SCryptPasswordEncoder;
 import org.springframework.security.web.authentication.UsernamePasswordAuthenticationFilter;
 import org.springframework.security.web.authentication.WebAuthenticationDetails;
+import org.springframework.security.web.authentication.www.BasicAuthenticationFilter;
 import org.springframework.security.web.context.request.async.WebAsyncManagerIntegrationFilter;
 import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
 import org.springframework.web.cors.CorsConfiguration;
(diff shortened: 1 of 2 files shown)


randomnicode:
Hmm, this seems okay to me, looks like there is an issue with the container being prepared for testing, will look into it

tamland:
@randomnicode I looked into the build failure a bit earlier (code here https://github.com/tamland/airsonic-advanced/commits/fix-docker-build) if you want some hints. The tests seems broken to me, i.e. depend on output of specific ffmpeg version, but I don't know..

randomnicode:
fixed here: https://github.com/airsonic-advanced/airsonic-advanced/pull/427

merge master and push again please?

the tests are tied to an ffmpeg version indeed, but the issue is that the container is now using focal whereas the ffmpeg ppa source we attach doesn't support focal for the latest ffmpeg build.

tamland:
@randomnicode passing now

airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
```diff
@@ -282,7 +283,9 @@ protected void configure(HttpSecurity http) throws Exception {
             http
                     .cors()
                     .and()
-                    .addFilterBefore(restAuthenticationFilter, UsernamePasswordAuthenticationFilter.class)
+                    .httpBasic()
+                    .and()
+                    .addFilterAfter(restAuthenticationFilter, BasicAuthenticationFilter.class)
```
randomnicode, line 288: Why is it necessary to have the rest auth filter now after the basic form auth filter?

tamland: Because it throws errors when the username/password parameters are missing and stops basic auth, [but it does handle existing authentication](https://github.com/airsonic-advanced/airsonic-advanced/blob/master/airsonic-main/src/main/java/org/airsonic/player/security/RESTRequestParameterProcessingFilter.java#L86). The comment even suggest it did work at one time so I just swapped them. Order shouldn't really matter as long as it allows the next filter to attempt authentication.

randomnicode:
@tamland going to merge this, since it's been here long enough and has had enough time for folks to comment on if it was an issue

PR #443, opened 2021-01-20
Provide ability to export podcasts to OPML format

randomnicode:
Mentioned here: https://github.com/airsonic/airsonic/issues/1422


alexdelorenzo:
Thanks!

Issue #447, opened 2021-01-21
search does not return correct albums/artists


cascooscuro:
## Problem description
I have seen  that search function does not  return albums/artists correctly in some cases. It happens both when I use web interface or though ISub Client.

For example, in this case I'm searching for the group "The La's":
![image](https://user-images.githubusercontent.com/5746813/105359558-e0b5e980-5bf7-11eb-97a9-d2fff56bd617.png)

The folder exists but search does not find it.

Another case. I know an Artists ("Zs") that has an album called "Noth". If I search by "Zs" it correctly finds the artist and it's album:
![image](https://user-images.githubusercontent.com/5746813/105359931-4e621580-5bf8-11eb-9cb0-de0381fd8406.png)

But if try to find by the name of the Album: "Noth", it doesn't find it:
![image](https://user-images.githubusercontent.com/5746813/105360315-bdd80500-5bf8-11eb-84e8-9a04e593752a.png)

 

## System information

Linux Arch
Version | 11.0.0-SNAPSHOT.20210121090405 – January 21, 2021 at 10:04:05 AM CET
Server | Apache Tomcat/9.0.41, java 11.0.10, Linux (1.08 GB / 1.29 GB)
nginx/1.19.6
Somewhat big library:
![image](https://user-images.githubusercontent.com/5746813/105362010-a7cb4400-5bfa-11eb-9006-e1b58294892f.png)

##Logs
A few errors but not sure if related:
```


2021-01-21 15:03:23.612 ERROR --- o.a.p.s.LoggingExceptionResolver         : x.x.x.x: An exception occurred while loading http://xxx.xxx.xxx/airsonica/websocket/846/2u23g2cj/eventsource

org.springframework.http.converter.HttpMessageNotWritableException: No converter for [class java.util.LinkedHashMap] with preset Content-Type 'text/event-stream;charset=UTF-8'
        at org.springframework.web.servlet.mvc.method.annotation.AbstractMessageConverterMethodProcessor.writeWithMessageConverters(AbstractMessageConverterMethodProcessor.java:299) ~[spring-webmvc-5.3.3.jar!/:5.3.3]
        at org.springframework.web.servlet.mvc.method.annotation.HttpEntityMethodProcessor.handleReturnValue(HttpEntityMethodProcessor.java:219) ~[spring-webmvc-5.3.3.jar!/:5.3.3]
[...]

PR #451, opened 2021-01-22
Use id3 index for searches also

randomnicode:
Fixes https://github.com/airsonic-advanced/airsonic-advanced/issues/447


randomnicode:
This PR will look through both id3 and folder based searches

PR #461, opened 2021-01-30
Fix getArtist endpoint starred data

randomnicode:
Originally pointed by and fixes: https://github.com/airsonic/airsonic/issues/1840

Retrieving media file data (folder artist) instead of artist id3 data


ctedgar (approved):
Hello @randomnicode! Looks like this addresses the issue 👍 (it was trying to fetch the starred information for artists from the mediafile/song starred table). 
This change brings it inline w/ createJaxbChild and createJaxbAlbum methods (all fetching starred information from their respective datastores)! 

Thanks for the quick turn around! 
Cris

PR #463, opened 2021-02-02
Test multiplatform release

randomnicode:
A test merge...


randomnicode:
Most of the work done in #438 albeit in PRs.

This mostly just automates the edge release process ONLY.

Note that now there is a redundancy in the container building process. The linux container gets built twice:
- Unit tests
- Linux Image built with docker-maven-plugin
- Integration tests run against image above
- Image(s) for various platform including Linux built and pushed

randomnicode:
The test removal as part of this PR will be removed separately later.

PR #482, opened 2021-02-15
Add mod transcoding to mp3 and install xmp in docker container to decode mod files

randomnicode:
Will need to register the following extensions as Music types in Settings -> General also

alm 669 mdl far xm mod fnk imf it liq wow mtm ptm rtm stm s3m ult dmf dbm med okt emod sfx m15 mtn amf gdm stx gmc psm j2b umx amd rad hsc flx gtk mgt mtp

Default adds these already, but won't override if a custom list already exists in the parameter file, so do add them.

Note, needs xmp. In Docker container, xmp is installed.

Fixes https://github.com/airsonic/airsonic/issues/1698


randomnicode:
Fixes https://github.com/airsonic/airsonic/issues/1698

PR #484, opened 2021-02-16
Order Albums by a secondary index other than create_date as create_dates are not unique

### LINKED ISSUE #483: Duplicate entries whilst paging through getAlbumList2 API responses
## Problem description

*Hello, I'm seeing duplicate album entries whilst paging through [getAlbumList2 ](http://www.subsonic.org/pages/api.jsp#getAlbumList2) API responses, in particular, when paging through the "newest" albums. E.g. I execute the request to get the first 20 newest albums (size=20 and offset=0). I then execute the request to get the next 20 newest albums (size=20 and offset=20) and item 18 from the first request appears in the second request.

I've snooped around the code base and it seems this is partly to do with how the "created" field of an album is being populated when the file is scanned (many albums have the exact same created time - I would think this would be highly unlikely given the timestamp should be precise enough). 

Another reason is due to the way the query to retrieve the albums from the datastore is executed. Looking at the [code to retrieve the newest albums](https://github.com/randomnicode/airsonic-advanced/blob/3a3edfc20b1a26a741d34f6ee115b7e2b520518d/airsonic-main/src/main/java/org/airsonic/player/dao/AlbumDao.java#L248),  we are ordering solely by the "created" field. For albums with the exact same "created" field, we have no guarantees for ordering (at least I'm not seeing any...). I think adding "id" to the order by would solve this as it's unique (see screenshots below). I have only noticed this in the getAlbumList responses but I suspect this can happen in any of the pageable API request where ordering by a timestamp field is an option.

Addressing either one or both would fix this - with the latter (adding "id" to the order by) being the most straight forward.

### Steps to reproduce

Connect to your airsonic DB 

1. Execute the query that airsonic uses to obtain the newest albums with a size of 20 and offset of 0
image

2. Execute the query that airsonic uses to obtain the newest albums with a size of 20 and offset of 20
image

Note how ordering is not guaranteed as there is a handful of albums with a "created" timestamp of "04:45:31:000".

Adding id to the order by resolves the issue by guaranteeing order
Running same query with id in order by, size 20 and offset 0
image

Running same query with id in order by, size 20 and offset 20
image

## System information

 * **Airsonic version**: *11.0.0-SNAPSHOT.20210130073922*
 * **Operating system**: *Linux (Raspbian)*
 * **Java version**: *java 11.0.9.1*
 * **Server**: *Apache Tomcat/9.0.31*
 * **Database**:*Postgres*
 * **Client**: *Postman*
 * **Language**: *English*

randomnicode:
Fixes #483

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/dao/AlbumDao.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/dao/AlbumDao.java
@@ -254,7 +254,7 @@ public class AlbumDao extends AbstractDao {
         args.put("count", count);
         args.put("offset", offset);
         return namedQuery("select " + QUERY_COLUMNS + " from album where present and folder_id in (:folders) " +
-                          "order by created desc limit :count offset :offset", rowMapper, args);
+                "order by created desc, id desc limit :count offset :offset", rowMapper, args);
     }
 
     /**


ctedgar (approved):
Hi @randomnicode, 
Thanks for this, this looks good! Maybe I'll raise a PR to address other instances of this. (basically add id to the order bys).

randomnicode:
That sounds good

PR #485, opened 2021-02-16
Adding unique secondary index to order bys

ctedgar:
Ensuring pageable api calls provide responses with consistent ordering


ctedgar:
This PR is an extension of #483 #484

Issue #511, opened 2021-04-12
Wrong cover picture displayed


SubZeroPL:
## Problem description

When we have mp3 file with multiple images embedded into tag the application always displays first one, which does not have to be front cover (i assume we want to display front cover image). Order of embedded images is not standardised, but each image has assigned it's type (see: [ID3 tag format](https://id3.org/id3v2.4.0-frames) pt. 4.14).
I did not check with other formats (flac, wma, ogg) as i mostly use mp3, but judging by docs from [Jaudiotagger PictureTypes class](http://www.jthink.net/jaudiotagger/javadoc/org/jaudiotagger/tag/reference/PictureTypes.html) it is handled the same.

### Steps to reproduce

1. Add new media folder with mp3 files containing multiple cover images in tag
2. Rescan if needed
3. Go to Home -> newly added album
4. First image from tag is displayed

Attached sample file (here front cover is image no.8)
[sample.zip](https://github.com/airsonic-advanced/airsonic-advanced/files/6296005/sample.zip)

## System information

 * **Airsonic version**: 11.0.0-SNAPSHOT.20210309214438
 * **Operating system**: Docker on Windows 10
 * **Java version**: included in docker image (Java 11 i think)
 * **Proxy server**: None
 * **Client**: Firefox 87, Ultrasonic 2.18.0
 * **Language**: english

## Additional notes
I see two possible solutions:
- display always front cover image based on PictureTypes (my current implementation)
- add configuration option to display image type selected by user (with fallback to front cover/first image from tag)

randomnicode:
Can we test this now and close the ticket if fixed?

SubZeroPL:
Tested with latest docker image, works as intended so issue can be closed.

randomnicode:
Thank you for your contribution!

PR #512, opened 2021-04-12
Change cover display for album

SubZeroPL:
Change cover display for album so we do not display always first image from file tag but image that is marked as front cover.

Proposed fix for #511.


randomnicode:
external db issues resolved in #523

PR #547, opened 2021-06-21
Strip terminating and embedded \0 in JaudioTaggerParser

Yetangitu:
Jaudiotagger sometimes returns strings with trailing and (rarely) embedded \0 characters, strip these before storing them in the database. This avoids problems with Postgresql (which can not store \0 in string data types) and it is more correct as these \0 characters should not be there in the first place. Strangely enough Jaudiotagger *does* strip terminating \0 characters when storing WAV info tags in ID3 fields, this patch extends this functionality to all other (string) types and also takes care of embedded stray \0's.

This patch is needed to make airsonic (any version) work reliably with Postgresql. It is a continuation of a similar PR (https://github.com/airsonic/airsonic/pull/1236) to the original airsonic which never got merged.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
@@ -131,9 +131,17 @@ public class JaudiotaggerParser extends MetaDataParser {
 
     private static String getTagField(Tag tag, FieldKey fieldKey) {
         try {
-            return StringUtils.trimToNull(tag.getFirst(fieldKey));
+            return stripNullCharacters(StringUtils.trimToNull(tag.getFirst(fieldKey)));
         } catch (Exception x) {
             // Ignored.
+            return null;
+        }
+    }
+
+    private String stripNullCharacters(String s) {
+        if (s != null) {
+            return s.replace("\0", "");
+        } else {
             return null;
         }
     }
--- a/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
@@ -131,17 +131,9 @@ public class JaudiotaggerParser extends MetaDataParser {
 
     private static String getTagField(Tag tag, FieldKey fieldKey) {
         try {
-            return stripNullCharacters(StringUtils.trimToNull(tag.getFirst(fieldKey)));
+            return StringUtils.replace(StringUtils.trimToNull(tag.getFirst(fieldKey)), "\0", " ");
         } catch (Exception x) {
             // Ignored.
-            return null;
-        }
-    }
-
-    private String stripNullCharacters(String s) {
-        if (s != null) {
-            return s.replace("\0", "");
-        } else {
             return null;
         }
     }


randomnicode:
Should we substitute /0 with a space or a semi colon character instead of just completely stripping it out?

Yetangitu:
Either is possible, realistically speaking it doesn't matter much since there hardy any files with embedded zeroes in tags. The few which are there ruin your day should you use Postgresql (as I do), hence this change. The "standard" (as far as it goes) seems to allow embedded zeroes, it does not state what to do with the resulting list. In an earlier version of this change I had included an option to replace them with a character of choice but this did not survive the PR discussion.

I'd say a space character is probably the least intrusive replacement given the lack of direction on how to interpret tags with embedded zeroes.

airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
```diff
@@ -131,13 +131,21 @@ public MetaData getRawMetaData(Path file) {
...
             return null;
         }
     }
 
+    private String stripNullCharacters(String s) {
+        if (s != null) {
+            return s.replace("\0", "");
+        } else {
+            return null;
+        }
```
randomnicode, line 146: Can just use StringUtils.replace(s, "\0", " ");
It'll do all the null checks for you.

Yetangitu: When using StringUtils.replace() there is no need for a separate method:

```
--- a/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
@@ -131,21 +131,13 @@ public class JaudiotaggerParser extends MetaDataParser {
 
     private static String getTagField(Tag tag, FieldKey fieldKey) {
         try {
-            return stripNullCharacters(StringUtils.trimToNull(tag.getFirst(fieldKey)));
+            return StringUtils.replace(StringUtils.trimToNull(tag.getFirst(fieldKey)), "\0", " ");
         } catch (Exception x) {
             // Ignored.
             return null;
         }
     }
 
-    private String stripNullCharacters(String s) {
-        if (s != null) {
-            return s.replace("\0", "");
-        } else {
-            return null;
-        }
-    }
-
     /**
      * Returns all tags supported by id3v1.
      */
```
This retains the original behaviour of returning null for empty tags by first using `trimToNull()`.

airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
```diff
@@ -131,13 +131,21 @@ public MetaData getRawMetaData(Path file) {
...
         try {
-            return StringUtils.trimToNull(tag.getFirst(fieldKey));
+            return stripNullCharacters(StringUtils.trimToNull(tag.getFirst(fieldKey)));
         } catch (Exception x) {
             // Ignored.
             return null;
         }
     }
 
+    private String stripNullCharacters(String s) {
```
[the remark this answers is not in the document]
Yetangitu: See below...

airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
randomnicode: Yup, that'll work, want to push a commit?

Yetangitu:
Voila

Issue #517, opened 2021-04-21
Unable playing music through DLNA because exception is thrown


Dikakus:
## Problem description

When I',m trying to play a song in a dlna client, an exception is thrown and nothing more happens. Nevertheless, I can play the same song if I use a browser (Edge for example) using the web client.

But if I use the same url that failed in the dlna client in the browser the excepction is thrown again. The url was catched from the DLNA client properties.

### Steps to reproduce

1. Access to a DLNA client.
2. Browse folders.
3. Select a song.

## System information

 * **Airsonic version**: 11.0.0-SNAPSHOT.20210309214438
 * **Operating system**: Raspbian 10 (burster)
 * **Java version**: 11.0.9.1
 * **Proxy server**: lighttp 1.4.53
 * **Client**: Philips TV, VLC, Windows 10 Player
 * **Language**: ES

[airsonic-log.txt](https://github.com/airsonic-advanced/airsonic-advanced/files/6353420/airsonic-log.txt)

randomnicode:
Your DB seems to have some integrity violations specifically around the player that was created for the DLNA client. You can:
(1) Delete your DB and rescan and try again (delete the airsonic-data/db folder)
(2) Try fixing your DB externally by yourself by finding the offending row in the Players table and deleting it. Alternatively, post your db folder here and maybe someone can take a look and fix it
(3) Try going to the Players section in your Settings page (if you can) and clone a new player and/or delete the offending player

Dikakus:
Thank you for your replay and your suggestions @randomnicode. As it could be a PLAYER(ID) issue, I tried to delete all my player through the web, but I can't create it anymore.

![imagen_2021-05-14_133314](https://user-images.githubusercontent.com/14204887/118264968-d5454180-b4b8-11eb-8db8-6f8bb9f125eb.png)

I tried to play a song from subsonic player in my phone and it throws the same exception. It seems that it can't create a new player. 

Any idea?

Dikakus:
Finally, I exported all my playlists, deleted the DB, reescanned my music, e imported the playlists. Unfortunatelly, I had two news issues: first, I lost my starred songs (I will restar it again) and second I still can't play DLNA songs with my TV. The log says:

```
2021-05-25 10:05:08.933  INFO --- o.a.p.c.StreamController                 : 192.168.0.36: anonymous listening to ... (etc).
...
[...]

PR #557, opened 2021-06-30
cache eviction bug for null entries

randomnicode:


randomnicode:
Should fix a part of the SpEL issue in #517

PR #576, opened 2021-07-21
Enabled sorting for playlists

SubZeroPL:
Minimal change to configuration of table displaying playlist contents.

### DIFF
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/playlist.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/playlist.jsp
@@ -20,7 +20,6 @@
                 stateDuration: 60 * 60 * 24 * 365,
                 ordering: true,
                 order: [],
-                orderFixed: [ 0, 'asc' ],
                 orderMulti: false,
                 pageLength: ${model.initialPaginationSize},
               <c:set var="paginationaddition" value="${fn:contains(' 10 20 50 100 -1', ' '.concat(model.initialPaginationSize)) ? '' : ', '.concat(model.initialPaginationSize)}" />
@@ -44,7 +43,7 @@
                     callback({data: songs});
                 },
                 stripeClasses: ["bgcolor2", "bgcolor1"],
-                columnDefs: [{ targets: "_all", orderable: false }],
+                columnDefs: [{ targets: "_all", orderable: true }],
                 columns: [
                     { data: "seq", className: "detail fit", visible: true },
                     { data: "starred",


randomnicode:
This doesn't actually change the playlist, only its view. You need to do an API call to save the playlist with the new order

SubZeroPL:
You're right, I've added saving. I don't know if it's best solution but it's working as I want it (playlist can be sorted by clicking on column headers or by dragging files around).

airsonic-main/src/main/webapp/WEB-INF/jsp/playlist.jsp
```diff
@@ -44,7 +45,7 @@
                     callback({data: songs});
                 },
                 stripeClasses: ["bgcolor2", "bgcolor1"],
-                columnDefs: [{ targets: "_all", orderable: false }],
+                columnDefs: [{ targets: "_all", orderable: true }],
```
randomnicode, line 48: should this be orderable if edits are not allowed by the role?

airsonic-main/src/main/webapp/WEB-INF/jsp/playlist.jsp
```diff
@@ -285,6 +286,16 @@
                     playlistMusicTable.one( "draw", function () {
                         onRearrange(playlistMusicTable.rows().indexes().toArray());
                     });
+                    playlistMusicTable.order([0, "asc"]);
+                }
+            });
+            playlistMusicTable.on("order", (e, settings, ordArr) => {
+                if (previousSortingOrder.col !== ordArr[0].col || previousSortingOrder.dir !== ordArr[0].dir) {
```
randomnicode, line 293: this seems lacking on the secondary (and further) sort order checks.

if you order by artist desc and then do an additional sort order on album asc, it'll save on the first order, but fail on the second order (because the first order remains identical)

also why are the checks necessary? does it save a call to the backend every time a draw is done (which would come back via websocket notification and cause another redraw, and then another order event, thus causing a loop)?

maybe just save the entire array as previousSortingOrder and deep compare the arrays?

airsonic-main/src/main/webapp/WEB-INF/jsp/playlist.jsp
```diff
@@ -285,6 +286,16 @@
...
+                    playlistMusicTable.order([ordArr[0].col, ordArr[0].dir]);
```
randomnicode, line 294: worried about a loop here.

would calling the 'order()' function within the 'order' event cause another 'order' event to be generated thus causing the loop?

airsonic-main/src/main/webapp/WEB-INF/jsp/playlist.jsp
SubZeroPL: you are right, fixed

airsonic-main/src/main/webapp/WEB-INF/jsp/playlist.jsp
SubZeroPL: that's exactly the reason

airsonic-main/src/main/webapp/WEB-INF/jsp/playlist.jsp
SubZeroPL: that's why i check not only column used for sorting but also sorting order, to enable multiple sort operations on the same column.
the check was necessary because 'order' event is raised on every table load, and table is reloaded after save so we had an endless loop without it

Issue #604, opened 2021-08-23
clean install running under tomcat 9 starts once, but fails second time


retropc:
first run is fine

once the server is shut down and restarted the server tries to load the hsqldb driver and fails (because it's presumably not in the war):

it then determines that this means the database should be upgraded, which also attempts to load the driver, and also fails

at this point the startup terminates

```
tomcat9[13458]: 2021-08-23 00:00:45.035  WARN --- o.a.p.u.LegacyHsqlMigrationUtil          : HSQLDB database driver version cannot be determined. Will make a copy of the DB prior to connecting. This may upgrade the database from version 2.5.0
tomcat9[13458]: java.sql.SQLException: No suitable driver
tomcat9[13458]:         at java.sql/java.sql.DriverManager.getDriver(DriverManager.java:298) ~[java.sql:na]
tomcat9[13458]:         at org.airsonic.player.util.LegacyHsqlMigrationUtil.isHsqlDbBackupNeeded(LegacyHsqlMigrationUtil.java:72) ~[classes/:11.0.0-SNAPSHOT]
tomcat9[13458]:         at org.airsonic.player.util.LegacyHsqlMigrationUtil.upgradeFileHsqlDbIfNeeded(LegacyHsqlMigrationUtil.java:148) ~[classes/:11.0.0-SNAPSHOT]
tomcat9[13458]:         at org.airsonic.player.spring.StartupEventListener.onApplicationEvent(StartupEventListener.java:17) ~[classes/:11.0.0-SNAPSHOT]
tomcat9[13458]:         at org.airsonic.player.spring.StartupEventListener.onApplicationEvent(StartupEventListener.java:1) ~[classes/:11.0.0-SNAPSHOT]
tomcat9[13458]:         at org.springframework.context.event.SimpleApplicationEventMulticaster.doInvokeListener(SimpleApplicationEventMulticaster.java:176) ~[spring-context-5.3.8.jar:5.3.8]
tomcat9[13458]:         at org.springframework.context.event.SimpleApplicationEventMulticaster.invokeListener(SimpleApplicationEventMulticaster.java:169) ~[spring-context-5.3.8.jar:5.3.8]
tomcat9[13458]:         at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:143) ~[spring-context-5.3.8.jar:5.3.8]
[...]

PR #605, opened 2021-08-23
fix 2nd startup under tomcat

retropc:
fixes issue #604

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/util/LegacyHsqlMigrationUtil.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/util/LegacyHsqlMigrationUtil.java
@@ -5,6 +5,7 @@ import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


airsonic-main/src/main/java/org/airsonic/player/util/LegacyHsqlMigrationUtil.java
```diff
@@ -76,8 +76,8 @@ public static boolean isHsqlDbBackupNeeded(String dbPath, String jdbcUrl) {
                 return true;
             }
         } catch (SQLException e) {
-            LOG.warn("HSQLDB database driver version cannot be determined. Will make a copy of the DB prior to connecting. This may upgrade the database from version {}", currentVersion, e);
-            return true;
+            LOG.warn("HSQLDB database driver version cannot be determined.", e);
+            return false;
```
randomnicode, line 80: This just stops it from making a backup.

This won't do anything to try and upgrade (or prevent it) any legacy dbs

retropc:
okay so this is a better way of doing it

I've dug through the spring-datasource loading carefully and this is what it was invoking to load the driver

this works in standalone mode and also under tomcat

PR #659, opened 2021-11-04
Relative paths

randomnicode:
This significantly overhauls how paths are stored in the backend in media_files

There is more discussion about it on https://github.com/airsonic/airsonic/issues/165

Previously:
 - media files had unique full paths
 - any cross join across tables was on basis of full paths
 - led to issues when migrating or moving the location of the music folder elsewhere.
   - it would consider the music folder a brand new folder and create new media files with new paths
   - led to loss of statistics per media file as well as accumulated cruft in db.
 - nested music folders aren't a problem
 - podcast folders were not considered part of music folders

Now:
 - media files now have a folder_id and a relative path.
 - the full path is resolved by joining the folder path with the relative path.
 - if the folder path changes, the media file remains unaware and when resolving the full path, utilizes the new folder path, no new db entries are created, and the old data remains
 - ~there is now a path restriction on music folders so they don't overlap. this is the cost of relative paths. overlapping causes ambiguity in how a particular file may be resolved (using folder 1 or using folder 2)~
 - music folders cannot be modified to be in their own hierarchy. if set to an ancestor, it would mean all the relative paths of the children would need to be recalibrated. if set to a descendant, some of the entries (previously children) would be orphaned.
   - instead, a new ancestral or a descendant folder can be created.
 - podcasts are now music folders
 - joins are mostly done on basis of media file ids instead of paths

THIS IS A NON-BACKWARDS COMPATIBLE CHANGE! BACK UP YOUR DATABASE PRIOR TO RUNNING THIS VERSION IN CASE YOU WANT TO REVERT!


Yetangitu:
When will this change be merged so I can push the new-and-improved CUE-support change?

randomnicode:
Need to address how to resolve the issue of adding a music folder that is directly in the hierarchy of an existing music folder.

Example:
Existing music folder: /a/b
How to add an ancestor: /a (make the scanner path broader)
How to add a descendant: /a/b/c (make the scanner path narrower)

Currently:
We won't let them add anything that is in the hierarchy. So the immediate issue is how can people broaden or narrow down the scope of the scanner. The only way to do this is to delete the existing folder in the hierarchy and then add the new one. However, this has consequences in that it will not retain the existing stats about the files under the folder, they will be marked as not present, and new duplicate ones will be scanned as part of the new folder.

To resolve:
No matter what, adding something in the hierarchy cannot add a new folder. You can expand or narrow an existing folder.
- Adding the same path: Denied immediately. This one never makes sense and should always be denied.
- Adding an ancestor: Can merge with the existing folder. Everything existing will already be in scope. Will need to adjust the relative paths of the files in the existing folder
- Adding a descendant: Can merge with the existing folder. Everything in scope needs to have its relative path readjusted. Everything that is out of scope suddenly should not be touched. It will be marked not present. Should it have its folder id changed?

Yetangitu:
>  Adding a descendant: Can merge with the existing folder. Everything in scope needs to have its relative path readjusted. Everything that is out of scope suddenly should not be touched. It will be marked not present. Should it have its folder id changed?

Depends on how the system should react when the parent folder (`/a/b`) is disabled. If the descendant (`/a/b/c`) should be disabled as well the `id` should stay the same, if it should remain visible it should be changed. The latter is what users will expect when they see separate entries in _Settings->Media folders_. This way descendant folders can also be used to hide certain parts of the hierarchy by marking them as disabled.

The same happens when adding an ancestor ( `/a`), what should happen if the ancestor is hidden? If the existing folder (`/a/b`) should remain visible the ancestor needs to have a different `id`.

So, all folders should have their own id, no matter whether they are in the tree. The scanner should check whether branches happen to be defined as a _Media folder_  and skip them if they do, they'll be scanned in their own thread.

When a descendant folder is removed from the _Media folders_ list its content should be merged with its ancestor (i.e. assign the ascendant folder `id` to the descendant's _MediaFile_s). If an ascendant is removed from the list nothing special needs to be done.

randomnicode:
> So, all folders should have their own id, no matter whether they are in the tree.

This is an interesting ambition.
Other than selective exclusions in the folder hierarchy, which are achieved by creating a new music folder in the hierarchy and then disabling it, what are the use cases for multiple folders in the same hierarchy?
It may create issues for addressing because now there can be multiple ways to address the same file. We can try such that the most specific folder always wins out, but it's not foolproof, especially with `..` and `.`. The adjustment of media_files would be difficult, because the cache for everything in scope needs to be invalidated (but we'd have to address this issue regardless). The user rights management would also need to be considered. For Folder 1: /a/b, Folder 2: /a/b/c, Folder 3: /a/b/c/d if a user has permissions for 3, upon creation of a Folder 4: /a/b/c/d/e/f, the user would suddenly lose access to the hierarchy under 4. Or deleting Folder 3 suddenly brings a user subject to the rights provided to Folder 2.

Currently, in this PR, as it stands, you can only have 1 folder in the hierarchy. You can either narrow its scope or broaden it, but not add another one in the hierarchy. This allows the mapping of (path) -> (folder_id, relative_path) to remain 1:1

The music folder check in the scanner isn't so easy, because there is no "own thread". It's a shared work-stealing thread pool, though this change may still be done.

| Action | One-folder per hierarchy | Multiple folders per hierarchy |
|---|---|---|
| Add folder with same path | Denied | Denied |
| Add ascendant folder | Delete descendant folder, reassign folder_id and relative_paths for descendant media_files | Just add |
| Add descendant folder | Delete ascendant folder, reassign folder_id and relative_paths for all files of ascendant that are now in scope of new folder, do not touch the rest, they will have a folder id that points to nothing | Reassign folder_id and relative_paths for all files of the _nearest ascendant_ that are in scope of new folder |
| Delete ascendant folder | Same as adding descendant folder | Just delete |
| Delete descendant folder | Same as adding ascendant folder | Reassign all files to the next highest ascendant |

Thoughts?

Yetangitu:
Use cases? Who needs use cases? In the current setup I don't see that many outside of the "blacklist folders already mentioned, I reacted mostly to the question of how to solve the problem rather than _why to solve it_, assuming the latter to be a solved problem - users want to be able to do this for `$reasons`. The reason for using a separate folder id for all folders no matter their location in the tree are the ones given, it makes it possible to enable/disable folders at will, e.g. to give certain users access to `/a/b/` but deny access to `/a//b/c` while others get the whole thing (think 'keep the children away from those scary/raunchy/otherwise_unsuitable audiobooks which are nested inside the _/mnt/audio/books_ directory', etc)

Skipping files which belong to a descendant folder is easy, just add a filter to `MediaScannerService.scanFile()` which skips _directories_ with an absolute path starting with a descendant path, e.g.:

``` java
    private void scanFile(MediaFile file, MusicFolder musicFolder, MediaLibraryStatistics statistics,
            Map<String, AtomicInteger> albumCount, Map<String, Artist> artists, Map<String, Album> albums,
            Map<Integer, Album> albumsInDb, Genres genres, Map<String, Boolean> encountered) {

        if (file.isDirectory() && settingsService.getAllMusicFolders(true,false).stream()
                .filter(p -> !p.equals(musicFolder.getPath()))
                .anyMatch(p -> file.getFullPath(musicFolder.getPath()).startsWith(p))) {
            LOG.trace("skipping directory {} (belongs to descendant music folder)", file);
            return;
        }

        if (scanCount.incrementAndGet() % 250 == 0) {
            broadcastScanStatus();
            LOG.info("Scanned media library with {} entries.", scanCount.get());
        }
        ...
        ...
        ...
```
Not tested, it could do with some optimisation but it should work unless I'm overlooking something.

randomnicode:
> Not tested, it could do with some optimisation but it should work unless I'm overlooking something.

Yes, actually, you're missing a very major point in the architecture here.

The issue is that you think media files are populated in the db at the media scanner level, so by circumventing the media scanner from seeing it, you'll prevent their population in the db, but that's not how it's designed at all. The media scanner actually gets notified post fact of the files in the db, and is mostly just a tool to index the files. By the time this check you're installing is executed, it's too late, the media file is already in the db under a different parent/folder id. All your check is doing is circumventing stuff from being indexed.

Here's a detailed example. Consider the call hierarchy for a new db scan for:
/a
  |-b
  |-c
where Folder A: /a and Folder B: /a/b

- MediaFileService.getMediaFile("", FolderA) gets called
  - Nothing in the DB, so MediaFileService.createMediaFile("", FolderA) gets called and creates an entry MediaFileA
- MediaFileScanner.scanFile(MediaFileA, FolderA...) gets called
  - the installed check in the code provided above passes and is skipped (not a descendant)
  - MediaFileService.getChildrenOf(MediaFileA...) gets called and now we get to the interesting part, which is where we think you might be mistaken
    - updateChildren(MediaFileA) gets called
      - Nothing in the DB: storedChildrenMap is empty
      - Walk through the children:
        - createMediaFile("b", FolderA) is called and creates MediaFileB1 <-- At this point your DB is already out of sync (this media file should've never been created or persisted)
        - createMediaFile("c", FolderA) is called and creates MediaFileC
        - updateMediaFile for the children and parent is called
    - we return a list of MediaFileB1 and MediaFileC
  - For MediaFileB1 and MediaFileC, we call MediaFileScanner.scanFile(MediaFileB1, FolderA...) and MediaFileScanner.scanFile(MediaFileC, FolderA...) respectively
    - MediaFileScanner.scanFile(MediaFileB1, FolderA...) runs afoul of your check, and terminates
    - MediaFileScanner.scanFile(MediaFileC, FolderA...) does not, and recurses from the second step of the parent
 - In parallel, MediaFileService.getMediaFile("", FolderB) gets called
   - Nothing in the DB, so MediaFileService.createMediaFile("", FolderB) gets called and creates an entry MediaFileB2 <-- Now you have two entries for the same folder
   - Continue on as before

At this point we have 2 ways of addressing /a/b: MediaFileB1 and MediaFileB2.

> Skipping files which belong to a descendant folder is easy, just add a filter to MediaScannerService.scanFile() which skips directories with an absolute path starting with a descendant folder path, e.g.:

No, not really. It's not easy, you can't really do it in MediaScanner.

Yetangitu:
Just put the test in `MediaFiles.getChildrenOf()`, when I first wrote this message it contained the option of putting the test either in scanFile or in the former. The effect is the same.

randomnicode:
> The effect is the same.

It isn't. As demonstrated by the example above. The former merely stops the indexing of a duplicate path, but it has already been created. The latter, is necessary, but is it sufficient? Are there any other paths that create a media file that might be a duplicate? Seems like createMediaFile is called in a bunch of other places?

> Why would anyone want to add a music folder which is already in the hierarchy and as such already part of the collection?

The answer was actually posed in the original manipulation of the question. The issue isn't so much as adding something that's already part of the collection, because you're right, that is indeed redundant. However, a user might want to reduce or narrow the scope of a music folder. He or she may not longer want the entirety of /a/ scanned, instead opting to scan /a/b only.
S/he can do they by (a) editing the existing folder, or (b) by adding a new folder in the hierarchy. In the original formulation, since the hierarchy is limited to one folder per hierarchy, the system needs to adjust in a way such that only one folder remains, whether by absorbing the new created folder or understanding that the user seeks to reduce the scope and modifying the existing folder accordingly.

> On the question about adding an ancestor the user can simply edit the path for the music folder instead of adding a new folder

Agreed, adding is a simple enough matter, until you realize that expanding the scope means your existing files are affected because their paths need adjusting also. Imagine a folder /a/b, and a media file in that folder with a relative path c/d.mp3. If you go and edit and expand the scope of the folder to /a now, the relative path of the existing media file needs to change with it and become b/c/d.mp3

Expansion and reduction of scope seem like symmetrical operations.

Yetangitu:
It should be sufficient from what I see, I've had more than enough acquaintance with `createMediaFile()` in making the CUE-patch to assume that it is. But... given that the original target was as narrow as allowing users to broaden or narrow the scope of an existing folder I think any solution is overkill - just let them edit the path and do a (full) scan which should mark files outside of scope as unavailable and pick up anything which came into scope after the edit. This is already handled by the code without the need for anything new as far as I know.

PR #663, opened 2021-11-10
More network headers

randomnicode:
Support for looking at X-Forwarded-For and X-Forwarded-Schema


nowheretobefound:
I think theres a typo in the header name in this PR , should be X-Forwarded-SchemE and not  X-Forwarded-SchemA

PR #704, opened 2021-12-12
Add Wavpack (wv) to the list of "Music files".

Yetangitu:
Add "wv" (Wavpack) to the list of "Music files", To actually get Airsonic to play these files to the web player the "wv" format will need to be added the "mp3 audio" transcoder.

### DIFF
--- a/airsonic-main/src/main/resources/liquibase/11.0/add-wavpack.xml
+++ b/airsonic-main/src/main/resources/liquibase/11.0/add-wavpack.xml
@@ -6,7 +6,10 @@
         <preConditions onFail="MARK_RAN">
             <sqlCheck expectedResult="0">select count(*) from transcoding where name='mp3 audio' and (source_formats like 'wv %' or source_formats like '% wv %' or source_formats like '% wv')</sqlCheck>
         </preConditions>
-        <sql>update transcoding set source_formats=((select source_formats from transcoding where name='mp3 audio') || ' wv') where name='mp3 audio'</sql>
+        <update tableName="transcoding">
+            <column name="source_formats" valueComputed="(concat((select source_formats from transcoding where name='mp3 audio'), ' wv'))" />
+            <where>name='mp3 audio'</where> 
+        </update>
         <rollback />
     </changeSet>
 </databaseChangeLog>
--- a/airsonic-main/src/main/resources/liquibase/11.0/add-wavpack.xml
+++ b/airsonic-main/src/main/resources/liquibase/11.0/add-wavpack.xml
@@ -1,0 +1,12 @@
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet id="add-wavpack" author="yetangitu">
+        <preConditions onFail="MARK_RAN">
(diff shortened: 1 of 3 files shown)


randomnicode:
> To actually get Airsonic to play these files to the web player the "wv" format will need to be added the "mp3 audio" transcoder.

Can we add that?

Yetangitu:
It can be added, yes. It can be added to the default transcoder, adding it to existing installations where admins may have edited the applicable formats takes more effort.

randomnicode:
I'd add it to just the default, since this change is only addressing defaults (maybe use a where value is original string)?

Yetangitu:
Or do this:

```sql
update transcoding set source_formats=((select source_formats from transcoding where name='mp3 audio') || ' wv') where name='mp3 audio';
```

With this precondition:

```sql
select count(*) from transcoding where name='mp3 audio' and (source_formats like '%wv %' or source_formats like '% wv' or source_formats like '% wv %');
```

i,e,:

```xml
<databaseChangeLog
        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
    <changeSet id="add-wavpack" author="yetangitu">
        <preConditions onFail="MARK_RAN">
            <sqlCheck expectedResult="0">select count(*) from transcoding where name='mp3 audio' and (source_formats like 'wv %' or source_formats like '% wv %' or source_formats like '% wv')</sqlCheck>
        </preConditions>
        <sql>update transcoding set source_formats=((select source_formats from transcoding where name='mp3 audio') || ' wv') where name='mp3 audio';</sql>
        <rollback />
    </changeSet>
</databaseChangeLog>
```
This works in all situations and gives consistent results for new installations which will have the `source_formats` set to `ogg oga aac m4a flac wav wma aif aiff ape mpc shn wv`.

randomnicode:
Yup, this is fine.

Two changes I'd make:
- Instead of a sql statement, I'd make it into a liquibase update xml node
- the concat operator || isn't global, you should use concat() instead

randomnicode:
there's a conflict here

Yetangitu:
Sigh - not any more

airsonic-main/src/main/resources/liquibase/11.0/add-wavpack.xml
```diff
@@ -0,0 +1,15 @@
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet id="add-wavpack" author="yetangitu">
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="0">select count(*) from transcoding where name='mp3 audio' and (source_formats like 'wv %' or source_formats like '% wv %' or source_formats like '% wv')</sqlCheck>
+        </preConditions>
+        <update tableName="transcoding">
+            <column name="source_formats" valueComputed="(concat((select source_formats from transcoding where name='mp3 audio'), ' wv'))" />
```
randomnicode, line 10: don't think you need all this.

You can probably just suffice with:
`valueComputed="(concat(source_formats, ' wv'))`

the where clause will take care of the selection

Yetangitu: True, I just cut and pasted the line from the raw SQL (which I much prefer).

Issue #698, opened 2021-12-10
Shuffle radio fails on PostgreSQL


nowheretobefound:
Got this when trying the shuffle radio :
Latest Airsonic on postgresql

There was an unexpected error (type=Internal Server Error, status=500).
org.springframework.jdbc.BadSqlGrammarException: PreparedStatementCallback; bad SQL grammar [select media_file.id, media_file.path, media_file.folder, media_file.type, media_file.format, media_file.title, media_file.album, media_file.artist, media_file.album_artist, media_file.disc_number, media_file.track_number, media_file.year, media_file.genre, media_file.bit_rate, media_file.variable_bit_rate, media_file.duration, media_file.file_size, media_file.width, media_file.height, media_file.cover_art_path, media_file.parent_path, media_file.play_count, media_file.last_played, media_file.comment, media_file.created, media_file.changed, media_file.last_scanned, media_file.children_last_updated, media_file.present, media_file.version, media_file.mb_release_id, media_file.mb_recording_id from media_file  where media_file.present and media_file.type = 'MUSIC' and media_file.folder in (?) and media_file.play_count >= ? order by rand() limit 100]; nested exception is org.postgresql.util.PSQLException: ERROR: function rand() does not exist
  Hint: No function matches the given name and argument types. You might need to add explicit type casts.
  Position: 843
	at org.springframework.jdbc.support.SQLStateSQLExceptionTranslator.doTranslate(SQLStateSQLExceptionTranslator.java:101)
	at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:70)
	at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:79)
	at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:79)
	at org.springframework.jdbc.core.JdbcTemplate.translateException(JdbcTemplate.java:1541)
	at org.springframework.jdbc.core.JdbcTemplate.execute(JdbcTemplate.java:667)
	at org.springframework.jdbc.core.JdbcTemplate.query(JdbcTemplate.java:713)
	at org.springframework.jdbc.core.JdbcTemplate.query(JdbcTemplate.java:738)
[...]

PR #707, opened 2021-12-13
Use database-specific random function

Yetangitu:
Fix for #698, "Shuffle radio fails on PostgreSQL"; use a database-specific random function.

PostgreSQL uses `random()`, Mysql/MariaDB and HSQLDB use `rand()`, use `MetaData.getDatabaseProductName()` to check which DBMS is used and adjust the function name accordingly. This clause will still fail on e.g. SQL Server, Oracle and DB2 since these use totally different constructions [1] to select random rows. If these DBMS are to be supported the query needs to be restructured.

 [1]
 - SQL Server uses `SELECT TOP 1 column FROM table ORDER BY NEWID ();`
 - Oracle uses `SELECT column FROM (SELECT column FROM table ORDER BY dbms_random.value) WHERE rownum = 1;`
 - DB2 uses `SELECT column RAND () as IDX  FROM table ORDER BY  IDX FETCH FIRST 1 ROWS ONLY;`

### DIFF
--- a/airsonic-main/src/main/resources/liquibase/11.0/changelog.xml
+++ b/airsonic-main/src/main/resources/liquibase/11.0/changelog.xml
@@ -18,4 +18,5 @@
     <include file="podcast-channel-rules.xml" relativeToChangelogFile="true"/>
     <include file="avatar-storage.xml" relativeToChangelogFile="true"/>
     <include file="move-user-table.xml" relativeToChangelogFile="true"/>
+    <include file="postgresql-add-rand-function.xml" relativeToChangelogFile="true"/>
 </databaseChangeLog>
--- a/airsonic-main/src/main/resources/liquibase/11.0/postgresql-add-rand-function.xml
+++ b/airsonic-main/src/main/resources/liquibase/11.0/postgresql-add-rand-function.xml
@@ -1,0 +1,15 @@
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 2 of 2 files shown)


Yetangitu:
A simple alias would work for databases which differ only in the name of the random function - PostgreSQL and SQLite - but it will fail for other DBMS. I don't expect many installations using Oracle or DB2 but SQL Server is more wide-spread (the Express and Developer version are free to download and use). Should these be supported as well?

Otherwise this would work:

```sql
create or replace function rand () returns double precision language sql as $$
    select random()
$$;
```

randomnicode:
i think until we have evidence people are using any of those, we can just make the change for postgres

Yetangitu:
Not tested, should work

randomnicode:
Can we add a test for the radio shuffle that exercises the rand() function?

Yetangitu:
What would that test do? The only difference between PostgreSQL random() and MySQL rand() is the name. they both produce pseudo-random double precision (PostgreSQL) or "floating point" (MySQL, precision is machine-dependent) numbers between 0 and 1.

randomnicode:
the test is at a higher level. the reason this bug went undetected was no one exercised that sql code in a test. testing that mediafileservice.getrandomsongs() does what it does would exercise the code

randomnicode:
we'll merge this pr, but without a test we'll need someone to test with postgres in an actual prod code.

Yetangitu:
That's me, I'm using PostgreSQL. It work. When I said "not tested" I meant I had not tested this specific version of the change since I've got a different build running which includes the previous iteration (with the "where" clause in the computedValue) and did not feel like building yet again with only that change. I tested the bare change in PostgreSQL, that works as well. Ergo, merge.

PR #738, opened 2022-01-14
Update jaudiotagger from 2.2.6.20210504 to 3.0.1

Yetangitu:
Update _jaudiotagger_ from `2.2.6.20210504` to `3.0.1`, see [CHANGES,txt](https://bitbucket.org/ijabz/jaudiotagger/src/master/CHANGES.txt) for what was changed between 2.2.5 (2.2.6 was never released) and this version.


randomnicode:
1. Have you tested this without any issues on local?

The reason this is a concern is because the existing version is a custom-built version checked into the local repo and likely was built with java 11. Moving it to this will probably downgrade the dependency to java 8 (since it seems the binary on maven is in java 8).

2. Perhaps add the clean up as part of the pr?
 - Remove the commented out custom bin repo in pom.xml since maven central now has the artifact
 - Delete the local unused binary (custom 2.2.6) that isn't being used anymore in the repo/ folder

The PR that added the binary is here, so you may use it as a guide for clean up:
https://github.com/airsonic-advanced/airsonic-advanced/pull/523/files

Yetangitu:
I built and tested with Java 11 (which is the version I'm running anyway), did a full scan over a 200.000 file collection without any problems so I don't expect problems there. I'll have a look at cleaning up the remains from the custom version one of the coming days.

PR #741, opened 2022-01-17
Fix DLNA browsing issue

randomnicode:
Fixes #740

Issue #740, opened 2022-01-16
DLNA Server:    java.lang.IllegalArgumentException: fromIndex(0) > toIndex(-1)


make-nz:
## Problem description

Connecting from several DLNA devices like Sony-TV, Yamaha Receiver,  Logitechmediaserver, Android(VLC)
Airsonic DLNA server only returns empty dirs/folders or "UPnP/DLNA request failed, please check the UPnP/DLNA server. "

### Steps to reproduce

1. Enable DLNA server in Airsonic
2. Test with clients, or devices you have
3. Go to About page(/help.view?) an check for UPnP error, followed by stack trace
    `UPnP error: java.lang.IllegalArgumentException: fromIndex(0) > toIndex(-1)`

## System information

Version | 11.0.0-SNAPSHOT.20220109073049 – January 9, 2022 at 8:30:49 AM CET [Commit: fa50bfd558b7a750f245ef6ef54ca4ee3852fc21]

Server | Apache Tomcat/9.0.48, java 11.0.13, FreeBSD (256.3 MB / 1.00 GB)

## Additional notes
`2022-01-16 16:42:39.841 INFO --- o.a.p.s.u.DispatchingContentDirectory : UPnP request - objectId: 0, browseFlag: BrowseDirectChildren, filter: *, firstResult: 0, maxResults: 0
2022-01-16 16:42:39.844 ERROR --- o.a.p.s.u.DispatchingContentDirectory : UPnP error: java.lang.IllegalArgumentException: fromIndex(0) > toIndex(-1)
java.lang.IllegalArgumentException: fromIndex(0) > toIndex(-1)
at java.base/java.util.AbstractList.subListRangeCheck(AbstractList.java:509) ~[na:na]
at java.base/java.util.ArrayList.subList(ArrayList.java:1138) ~[na:na]
at org.airsonic.player.util.Util.subList(Util.java:96) ~[classes!/:11.0.0-SNAPSHOT]
at org.airsonic.player.service.upnp.UpnpContentProcessor.browseRoot(UpnpContentProcessor.java:80) ~[classes!/:11.0.0-SNAPSHOT]
at org.airsonic.player.service.upnp.DispatchingContentDirectory.browse(DispatchingContentDirectory.java:121) ~[classes!/:11.0.0-SNAPSHOT]
[...]

PR #743, opened 2022-01-19
Fix DLNA overflows in long > int casts

randomnicode:
Further fixes for #740 and continuation of #741 particularly for comment  https://github.com/airsonic-advanced/airsonic-advanced/pull/741#issuecomment-1015482222

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/upnp/AlbumUpnpProcessor.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/upnp/AlbumUpnpProcessor.java
@@ -19,6 +19,7 @@
 */
 package org.airsonic.player.service.upnp;
 
+import com.google.common.primitives.Ints;
 import org.airsonic.player.dao.AlbumDao;
 import org.airsonic.player.dao.MediaFileDao;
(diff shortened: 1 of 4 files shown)


make-nz:
Thanks again

This has fixed all previous UPnP errors, as far as I can tell.

But browsing the folders results in blank/empty returns — not throwing any Java Exception, or any other messages on the server.
**This is regardless of what client I'm using!**

Here are my sources… 

```
NAME	SIZE (MB)	FILES	FOLDERS	SUBFOLDERS
\\A	 8.309,35	2.426	    176	        58



NAME	SIZE (MB)	FILES	FOLDERS	SUBFOLDERS
\\B	62.884,32	17.102	  1.068	        18



NAME	SIZE (MB)	FILES	FOLDERS	SUBFOLDERS
\\C    135.357,35	29.431	  3.304	     1.258
```

With just A and B enabled it is working. Enabling C causes the issue.
(when I mount this source (C) in Jellyfin or Gerbera it is not causing any issues)

Is the structure or size a problem here?
Is there any verbose logging option, to get to the bottom of the problem, maybe?

Issue #619, opened 2021-09-13
Feature Request - Setting for session timeout


Trufax:
Hi,

i figure if you forget to set the checkbox to stay logged in is very low. Sometimes i stops right during the playback of a Set and i have to reenter my credentials again (but i keeps starts playing than automatically which is good.

Would it be possible to have a setting for the session timeout ? Or is it even possible maybe via config and i just did not see it ?

Thanks

regards

PR #753, opened 2022-01-27
Add general setting to configure session timeout

NoItemName:
Enables session timeout configuration. This feature doesn't override _remember me_ checkbox.
Implements [#619](https://github.com/airsonic-advanced/airsonic-advanced/issues/619)


randomnicode:
Does this spring config not suffice?
`server.servlet.session.timeout=60s`
specified in https://docs.spring.io/spring-boot/docs/current/reference/html/application-properties.html#application-properties.server.server.servlet.session.timeout

It's a good idea to just use the spring property without overriding manually in a custom config class. okay to have a ui for specifying it. We will provide PR comments accordingly

airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
```diff
@@ -86,6 +86,7 @@
     private static final String KEY_WELCOME_SUBTITLE = "WelcomeSubtitle";
     private static final String KEY_WELCOME_MESSAGE = "WelcomeMessage2";
     private static final String KEY_LOGIN_MESSAGE = "LoginMessage";
+    private static final String KEY_SESSION_TIMEOUT_MESSAGE = "SessionTimeout";
```
randomnicode, line 89: Probably should be named `server.servlet.session.timeout` so it overrides the spring boot value

airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
```diff
@@ -188,6 +189,7 @@
             "\\\\ \\\\\n" +
             "To change or remove this message, log in with administrator rights and go to <a href='settings.view'>Settings</a> > <a href='generalSettings.view'>General</a>.";
     private static final String DEFAULT_LOGIN_MESSAGE = null;
+    private static final int DEFAULT_SESSION_TIMEOUT_MINUTES = 60;
```
randomnicode, line 192: Maybe a String or Duration?

org.springframework.boot.web.servlet.server.Session details the timeout field as a Duration

default is 30m it seems

airsonic-main/src/main/java/org/airsonic/player/spring/CustomSessionListener.java
```diff
@@ -0,0 +1,22 @@
...
+
+import org.airsonic.player.service.SettingsService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Configuration;
+
+import javax.servlet.http.HttpSessionEvent;
+import javax.servlet.http.HttpSessionListener;
+
+@Configuration
+public class CustomSessionListener implements HttpSessionListener {
```
randomnicode, line 11: this class isn't needed

airsonic-main/src/main/resources/org/airsonic/player/i18n/ResourceBundle_en.properties
```diff
@@ -373,6 +373,7 @@ generalsettings.welcometitle=Welcome title
 generalsettings.welcomesubtitle=Welcome subtitle
 generalsettings.welcomemessage=Welcome message
 generalsettings.loginmessage=Login message
+generalsettings.sessiontimeout=Session timeout in minutes. Changes will take effect on a new login
```
randomnicode, line 376: if it's a string, session timeout will be a duration string, if no postfix then in seconds.
also not sure when changes take effect with the spring boot var, maybe at application restart?

airsonic-main/src/main/java/org/airsonic/player/command/GeneralSettingsCommand.java
```diff
@@ -172,6 +173,14 @@ public void setLoginMessage(String loginMessage) {
         this.loginMessage = loginMessage;
     }
 
+    public Integer getSessionTimeout() {
```
randomnicode, line 176: string or duration?

randomnicode (changes requested):

NoItemName:
Thanks for the feedback, i will work on the changes!

randomnicode (approved):

PR #783, opened 2022-02-02
Fix multiple podcast downloads (fixes #774)

### LINKED ISSUE #774: Duplicated podcasts and downloading very slow
## Problem description

Despite the previous fix, and starting my podcasts database from scratch, I still have multiple copies of each file.

Also the time to download and process seems to be quite long, I have a reasonable fibre link.

### Steps to reproduce

Add a podcast feed and wait for duplicates to appear.

Test retrieving a podcast using wget, and it downloads in 15s, however Airsonic seems to take much longer.  Currently not yet completed the first podcast feed after nearly 24 hours, although the duplicates aren't helping.

Podcast feed currently downloading/tested - https://rss.acast.com/redhanded

wget output - 

Connecting to dm4p36fbs3hl0.cloudfront.net (dm4p36fbs3hl0.cloudfront.net)|52.84.93.74|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 83906642 (80M) [audio/mpeg]
Saving to: ‘8019453.mp3’

8019453.mp3                                     100%[=====================================================================================================>]  80.02M  5.62MB/s    in 15s     

2022-01-31 07:37:09 (5.51 MB/s) - ‘8019453.mp3’ saved [83906642/83906642]

## System information

Airsonic = 11.0.0-SNAPSHOT.20220129040247
OS = Ubuntu server 20.04.3

Yetangitu:
This change contains the following fixes:
 - make podcast downloads more robust in the face of concurrent channel refresh attempts, this fixes the relevant portion of #774 (Duplicated podcasts)
 - get rid of the annoying `InvalidAudioFrameException` log spam on every podcast download by delaying the creation of the episode MediaFile until the file has actually been downloaded
 - reset channel status to `COMPLETED` in case of interrupted channel refresh attempts

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/PodcastService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/PodcastService.java
@@ -542,8 +542,8 @@ public class PodcastService {
                 })
                 .filter(Objects::nonNull)
                 // Sort episode in reverse chronological order (newest first)
-                .sorted(Comparator.comparingLong(k -> k.getPublishDate() == null ? 0L : -k.getPublishDate().toEpochMilli()))
-                .forEach(episode -> {
+                .sorted(Comparator.comparing((PodcastEpisode episode) -> episode.getPublishDate()).reversed())
+                .forEachOrdered(episode -> {
                     if (counter.decrementAndGet() < 0) {
                         episode.setStatus(PodcastStatus.SKIPPED);
                     }
--- a/airsonic-main/src/main/java/org/airsonic/player/service/PodcastService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/PodcastService.java
@@ -37,6 +37,7 @@ import org.airsonic.player.util.Util;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang3.RandomStringUtils;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.http.Header;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.config.CookieSpecs;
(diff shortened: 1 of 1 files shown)


Yetangitu:
Ignore the failed check (CI failure).

airsonic-main/src/main/java/org/airsonic/player/service/PodcastService.java
```diff
@@ -662,6 +684,12 @@ private boolean isEpisodeDeleted(PodcastEpisode episode) {
         return episode == null || episode.getStatus() == PodcastStatus.DELETED;
     }
 
+    // include episode == null to avoid attempts to download episodes for deleted channels
```
randomnicode, line 687: this check should likely be added to isEpisodeDeleted

airsonic-main/src/main/java/org/airsonic/player/service/PodcastService.java
```diff
@@ -662,6 +684,12 @@ private boolean isEpisodeDeleted(PodcastEpisode episode) {
...
+    private boolean isEpisodeQueued(PodcastEpisode episode) {
```
randomnicode, line 688: rename to skipDownload?

also add PodcastStatus.SKIPPED?

airsonic-main/src/main/java/org/airsonic/player/service/PodcastService.java
Yetangitu: It is already part of isEpisodeDeleted:

```java
    private boolean isEpisodeDeleted(PodcastEpisode episode) {
        episode = podcastDao.getEpisode(episode.getId());
        return episode == null || episode.getStatus() == PodcastStatus.DELETED;
    }
```

airsonic-main/src/main/java/org/airsonic/player/service/PodcastService.java
Yetangitu: I've thought of several names, none of which really cover the true intention of this test which is both to avoid double downloads as well as to keep it from downloading deleted podcasts. `skipDownload` doesn't really get across the intention, it reeks of `PodcastStatus.SKIPPED` with which is has nothing to do - that is meant for episodes which have never been selected for downloading. Since episodes which are never selected for download do not reach this function `PodcastStatus.SKIPPED` does not need to be part of this test.

Yetangitu:
BTW, there is another problem in this download code which I have not yet fixed but could add to this PR: when AA is configured to download the last X episodes for a podcast, when adding a new podcast it downloads a random selection of X episodes from the list of available episodes instead of the most recent X, at least for some podcasts. It seems that the sorting done on line 545:

```java
                // Sort episode in reverse chronological order (newest first)
                .sorted(Comparator.comparingLong(k -> k.getPublishDate() == null ? 0L : -k.getPublishDate().toEpochMilli()))
                .forEach(episode -> {
                    if (counter.decrementAndGet() < 0) {
                        episode.setStatus(PodcastStatus.SKIPPED);
                    }
```
...doesn't work reliably (maybe due to a badly formatted or missing `pubDate`).

PR #804, opened 2022-02-07
Configurable search count

### LINKED ISSUE #748: Feature request: return ALL search results - don't limit them
the topic says it all.
Whenever a search is done and more than 25 results SHOULD be displayed, then the limit kicks in.
If i want to search something which i know has more than 25 results, i have to jump through hoops to get to a solution.
Please either show ALL results or make it configurable how many results are shown.

randomnicode:
Fixes #748

Setting is on a per-user basis


randomnicode:
Hi, this isn't the pr for the change described. That would be https://github.com/airsonic-advanced/airsonic-advanced/pull/801

Please open a new issue for this. Also add your folder structure.

PR #840, opened 2022-02-27
Album Artists in tables

randomnicode:


The-Exterminator:
This table makes no sense as both the album and the artist are there.
Can you under "advanced" do one can turn it off

randomnicode:
Artist != Album Artist != Album

Some albums are made by artists, but have guest or collaborating artists.

randomnicode:
We can add a feature to turn it off

The-Exterminator:
> We can add a feature to turn it off

You can made it Settings -> Personal ???

The-Exterminator:
> Artist != Album Artist != Album
> 
> Some albums are made by artists, but have guest or collaborating artists. Or think of soundtracks where a song may be sung by someone, but they're only a collaborator, the credit for the album goes to the composer (who is the album artist).

I have a lot of album, nd all tags are from mp3tags and i have never see it :-(

randomnicode:
It's a pretty standard ID3 tag.

The-Exterminator:
> It's a pretty standard ID3 tag.

All my mp3 have Discogs and MusicBrainz tags, but the program is [mp3tags](https://www.mp3tag.de/en/)

Issue #871, opened 2022-03-24
SQL Exception when cleaning up database, using external MySQL via JDBC


B16WLL:
## Problem description

I'm using Airsonic Advanced with an external MySQL DB. Everything works perfectly, web interface, DSub connection,folders scan, music plays, artwork displays etc.
Except for cleaning up the database.

The settings page returns a 500 server error with the following information. Luckily the SQL command that's causing the issue is listed at the top of the trace

Status | 500
-- | --
Error | Internal Server Error
Message |  
Path | /musicFolderSettings.view
Time | Thu Mar 24 18:19:43 GMT 2022
Exception |  
Java version | Ubuntu 11.0.14
Operating system | Linux 5.15.0-23-generic
Server | Apache Tomcat/9.0.48
Memory | Used 160 of 206 MB
[...]

PR #880, opened 2022-04-04
Modify CoverArtDao,java

B16WLL:
Proposed fix for issue #871

Guidelines for Contributing
---------------------------

Airsonic development is a community project, and contributions are welcomed. Here are a few guidelines you should follow before submitting:

  1.  **License Acceptance** All contributions must be licensed as [GNU GPLv3](https://github.com/airsonic/airsonic/blob/develop/LICENSE.txt) to be accepted. Use [`git commit --signoff`](https://jk.gs/git-commit.html) to acknowledge this.
  2.  **No Breakage** New features or changes to existing ones must not degrade the user experience. This means do not introduce bugs, remove functionality, or make large changes to existing themes/UI without prior discussion in an Issue.
  3.  **Coding standards** Language best-practices should be followed, comment generously, and avoid "clever" algorithms. Refactoring existing messes is great, but watch out for breakage.
  4.  **Be bold!** Without contributions, this project will vanish. If you just want to help out, try [submiting a patch](https://github.com/airsonic/airsonic/issues?q=is%3Aissue+is%3Aopen+label%3Apatches-welcome) for an open Issue.
  5.  **Stay relevant** Issues or commentary that is off-topic or tangential to Airsonic development is subject to moderation. Questions should be focused on improving documentation to solve a problem. Visit [Reddit](https://www.reddit.com/r/airsonic) or [IRC](http://webchat.freenode.net?channels=%23airsonic) for community discussion.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/dao/CoverArtDao.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/dao/CoverArtDao.java
@@ -51,9 +51,9 @@ public class CoverArtDao extends AbstractDao {
 
     @Transactional(isolation = Isolation.READ_COMMITTED)
     public void expunge() {
-        update("delete from cover_art where entity_type='MEDIA_FILE' and entity_id in (select ca.entity_id from cover_art ca left join media_file m on ca.entity_id=m.id where m.id is null and ca.entity_type='MEDIA_FILE')");
-        update("delete from cover_art where entity_type='ALBUM' and entity_id in (select ca.entity_id from cover_art ca left join album a on ca.entity_id=a.id where a.id is null and ca.entity_type='ALBUM')");
-        update("delete from cover_art where entity_type='ARTTIST' and entity_id in (select ca.entity_id from cover_art ca left join artist a on ca.entity_id=a.id where a.id is null and ca.entity_type='ARTIST')");
+        update("DELETE FROM cover_art WHERE entity_type='MEDIA_FILE' AND entity_id IN (SELECT media_file.id FROM media_file WHERE media_file.id IS NULL)");
+        update("DELETE FROM cover_art WHERE entity_type='ALBUM' AND entity_id IN (SELECT album.id FROM album WHERE album.id IS NULL)");
+        update("DELETE FROM cover_art WHERE entity_type='ARTIST' AND entity_id IN (SELECT artist.id FROM artist WHERE artist.id IS NULL)");
     }
 
     private static class CoverArtRowMapper implements RowMapper<CoverArt> {


airsonic-main/src/main/java/org/airsonic/player/dao/CoverArtDao.java
```diff
@@ -51,9 +51,9 @@ public void delete(EntityType type, int id) {
 
     @Transactional(isolation = Isolation.READ_COMMITTED)
     public void expunge() {
-        update("delete from cover_art where entity_type='MEDIA_FILE' and entity_id in (select ca.entity_id from cover_art ca left join media_file m on ca.entity_id=m.id where m.id is null and ca.entity_type='MEDIA_FILE')");
-        update("delete from cover_art where entity_type='ALBUM' and entity_id in (select ca.entity_id from cover_art ca left join album a on ca.entity_id=a.id where a.id is null and ca.entity_type='ALBUM')");
-        update("delete from cover_art where entity_type='ARTTIST' and entity_id in (select ca.entity_id from cover_art ca left join artist a on ca.entity_id=a.id where a.id is null and ca.entity_type='ARTIST')");
+        update("DELETE FROM cover_art WHERE entity_type='MEDIA_FILE' AND entity_id IN (SELECT media_file.id FROM media_file WHERE media_file.id IS NULL)");
```
randomnicode, line 54: this is incorrect and not equivalent

cover_art table has a column called entity_id which is, for entity_type='MEDIA_FILE', populated by ids that are rows in media_file table (media_file.id)

so imagine a media_file:
| id | path | ... |
|---|-------|---|
|1  | filea.mp3 | - |
|2 | fileb.mp3 | - |

with its cover_art
| entity_id |entity_type| ... |
|---|-------|---|
|1  | MEDIA_FILE | - |
|2  | MEDIA_FILE | - |

what ends up happening before this expunge is that we delete some of the media_files. now in this method we're looking to delete the cover_art for which no media_file exists

so imagine from our example earlier, if the row with id 1 gets deleted from media_files
| id | path | ... |
|---|-------|---|
|2 | fileb.mp3 | - |

and the cover_art is like 
| entity_id | type| ... |
|---|-------|---|
|1  | MEDIA_FILE | - |
|2 | MEDIA_FILE | - |

how do we delete this cover art?

- we left-join cover_art with media_file on ids, that'll yield:

| entity_id | type| id | path |
|----------|-------|---|------|
|1  | MEDIA_FILE | null | null |
|2 | MEDIA_FILE | 2 | fileb.mp3 |

and we can delete everything that has a null id.

- `SELECT media_file.id FROM media_file WHERE media_file.id IS NULL` will yield nothing because id is a primary key for media_file, it cannot be null

Issue #869, opened 2022-03-23
[Feature] JukeBox Bit Perfect Play


fangxx3863:
## Problem description

In the new version, Airsonic forces use ffmpeg to downsample HiRes audio to 44.1kHz. 
I wish there was an option to play the original file. So I tried to modify the code, unfortunately my code level is too poor to do this. 
I think we can call mpv to play files, mpv can use socket to communicate, it may be able to replace AudioPlayer.java
My English is not good, please forgive some mistakes,thanks!

### Steps to reproduce

1. Play a 96kHz/24bits audio 
2. ffmpeg will downsample to 44.1kHz/16bits and started to play

## System information

 * **Airsonic version**: *11.0.0-SNAPSHOT.20220318050825*
 * **Operating system**: *Debian with docker*
 * **Java version**: *14*
 * **Proxy server**: *None*
 * **Client**: *DSub*
 * **Language**: *Chinese*

randomnicode:
"Settings" -> "Transcoding tab" -> "Jukebox command" field describes the command used for jukebox play. You don't need to modify code. If you want the files to play with a different sample rate or just as they are, modify the command parameters accordingly.

fangxx3863:
> "Settings" -> "Transcoding tab" -> "Jukebox command" field describes the command used for jukebox play. You don't need to modify code. If you want the files to play with a different sample rate or just as they are, modify the command parameters accordingly.

![image](https://user-images.githubusercontent.com/48589001/162351912-d2d536fa-23cc-40cd-a836-23b9869288a8.png)

This line is dead written 
I tried modifying the code to extract audio parameters by regex matching the ffmpeg command line

PR #882, opened 2022-04-10
Fix JukeBox

fangxx3863:
Get audio information from transcoding commands. 
This PR can get sample rate, bits, channel from Jukebox transcode command. 
For example `ffmpeg -ss %o -i %s -map 0:0 -v 0 -ar 176400 -ac 2 -f s32be -` i use regular to match `-ar` `-ac` `-f`.  
In my system, it works well.

To fix my issue https://github.com/airsonic-advanced/airsonic-advanced/issues/869

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/jukebox/AudioPlayer.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/jukebox/AudioPlayer.java
@@ -25,12 +25,15 @@ import org.slf4j.LoggerFactory;
 
 import javax.sound.sampled.AudioFormat;
 import javax.sound.sampled.AudioSystem;
+import javax.sound.sampled.DataLine;
 import javax.sound.sampled.FloatControl;
 import javax.sound.sampled.SourceDataLine;
 
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static org.airsonic.player.service.jukebox.AudioPlayer.State.*;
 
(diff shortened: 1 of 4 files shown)


airsonic-main/src/main/java/org/airsonic/player/service/jukebox/AudioPlayer.java
```diff
@@ -49,16 +52,68 @@
...
-    public AudioPlayer(InputStream in, Listener listener) throws Exception {
+    public AudioPlayer(InputStream in, String command, Listener listener) throws Exception {
         this.in = in;
         this.listener = listener;
 
-        AudioFormat format = new AudioFormat(AudioFormat.Encoding.PCM_SIGNED, 44100.0F, 16, 2, 4, 44100.0F, true);
-        line = AudioSystem.getSourceDataLine(format);
+        // SampleRate
+        String sampleRate = "";
+        Pattern r1 = Pattern.compile("\\-ar (.*?)\\ ");
```
randomnicode, line 67: please compile these patterns outside the method so they're not compiled again and again?

airsonic-main/src/main/java/org/airsonic/player/service/jukebox/AudioPlayer.java
```diff
@@ -49,16 +52,68 @@
...
 
-    public AudioPlayer(InputStream in, Listener listener) throws Exception {
+    public AudioPlayer(InputStream in, String command, Listener listener) throws Exception {
         this.in = in;
         this.listener = listener;
 
-        AudioFormat format = new AudioFormat(AudioFormat.Encoding.PCM_SIGNED, 44100.0F, 16, 2, 4, 44100.0F, true);
-        line = AudioSystem.getSourceDataLine(format);
+        // SampleRate
+        String sampleRate = "";
```
randomnicode, line 66: set a default sample rate in case?

airsonic-main/src/main/java/org/airsonic/player/service/jukebox/AudioPlayer.java
```diff
@@ -49,16 +52,68 @@
...
+            pt = pt3.group(1);
+        }
+        LOG.debug("Format: " + pt);
+
+        // Bits
+        String bits = pt.substring(1, 3);
+        LOG.debug("Bits: " + bits);
+
+        // Big/Little Endian
+        Boolean sl = true;
```
randomnicode, line 97: unboxed

boolean instead of Boolean

airsonic-main/src/main/java/org/airsonic/player/service/jukebox/AudioPlayer.java
```diff
@@ -49,16 +52,68 @@
...
+        if (pt.substring(3, 5).equals("be")) {
+            sl = true;
+        }
```
randomnicode, line 100: this condition not needed.

only need condition when variable changes (starts as true)

fangxx3863:
First of all, thanks for your reply 
My code level is really bad, so please forgive me 
I also hesitated for a long time whether to submit such a bad code 
I will modify the code according to your comments

